PR #4, opened 2019-09-26
feat: using token creation tx information to display more data on tx detail screen

pedroferreira1:
image


obiyankenobi:
Can you add some screenshots?

pedroferreira1:
@obiyankenobi added the screenshot only for the page that is different from the wallet desktop.

Still waiting for the other PR for us to decide the changes and then I replicate here

msbrogli:
I think we should clearly show the `token id`. And, if it is not difficult, show the total number of transactions of that token.

src/App.js
```diff
@@ -16,7 +17,13 @@ import { dashboardUpdate, isVersionAllowedUpdate } from "./actions/index";
 import { connect } from "react-redux";
 import versionApi from './api/version';
 import helpers from './utils/helpers';
+import MemoryStore from './store';
```
msbrogli: Why don't we move the `MemoryStore` to the hathor-wallet-lib?

src/components/TxData.js
```diff
@@ -90,29 +102,23 @@ class TxData extends React.Component {
...
    *
-   * @param {Array} tokens Array of already added tokens
    * @param {number} tokenData Represents the index of the token in this transaction
+   * @return {Object} Token config object with {uid, name, symbol}
    */
-  checkToken = (tokens, tokenData) => {
-    if (tokenData === HATHOR_TOKEN_INDEX) {
+  checkToken = (tokenData) => {
+    if (tokenData === hathorLib.constants.HATHOR_TOKEN_INDEX) {
       return;
```
msbrogli: What is the return type of this method? It seems that there are some `return` along the method, but the last line is a `return tokenConfig`. Maybe we should `return null` here?

src/store.js
```diff
@@ -0,0 +1,32 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+export default class MemoryStore {
```
msbrogli: I think we should move it to the `hathor-walelt-lib`.

src/screens/TokenDetail.js
```diff
@@ -0,0 +1,207 @@
...
+      );
+    }
+
+    return (
+      <div className="content-wrapper flex align-items-center">
+        <div className='d-flex flex-row align-items-start justify-content-between token-detail-top'>
+          <div className='d-flex flex-column justify-content-between mt-4'>
+            <div className='token-wrapper d-flex flex-row align-items-center mb-3'>
+              <p className='token-name mb-0'>
+                <strong>{this.state.token.name} ({this.state.token.symbol})</strong>
```
msbrogli: I don't know where, but we should add a warning message that no one can trust the token without checking the token id. Also, the token id should be really visible.

I think we should also include a list of "conflicting" tokens to let users know about possible fraud.

src/App.js
pedroferreira1: In this case is another problem. The explorer does not need any information saved in the `localStorage`. We are only using some util methods. I think in a lib refactor we should make the storage optional.

src/store.js
pedroferreira1: Same as this https://github.com/HathorNetwork/hathor-explorer/pull/4#discussion_r334080351

src/screens/TokenDetail.js
pedroferreira1: The alert message was added and we have talked about creating this conflicting tokens alert as an issue to be discussed in the future

src/components/TxData.js
```diff
@@ -33,6 +33,9 @@ class TxData extends React.Component {
    */
   state = { raw: false, children: false, tokens: [] };
 
+  // Array of token uid that was already found to show the symbol
+  tokensFound = [];
```
obiyankenobi: Do we really need this? Now that the API returns all the tokens and their information, we should be able to just assemble a dict with this info (`Dict[uid, tokenConfig]`) and use it for getting the information. `tokens` on the state wouldn't be needed.

src/screens/TokenDetail.js
```diff
@@ -0,0 +1,216 @@
...
+  copied = (text, result) => {
+    if (result) {
+      // If copied with success
+      this.showSuccess('Configuration string copied to clipboard!');
+    }
+  }
+
+  /**
+   * Checks if the recently arrived transaction should trigger an update on the list
+   * It returns true if the transaction there is an input or output of the token
```
obiyankenobi: ... true if, in the transaction, there is an ...

src/screens/TokenDetail.js
```diff
@@ -0,0 +1,216 @@
...
+              {getShortConfigurationString()}
+              <CopyToClipboard text={configurationString} onCopy={this.copied}>
+                <i className="fa fa-clone pointer ml-1" title="Copy to clipboard"></i>
+              </CopyToClipboard>
+            </span> 
+            <QRCode size={200} value={configurationString} />
+            <a className="mt-2" onClick={(e) => this.downloadQrCode(e)} download={`${this.state.token.name} (${this.state.token.symbol}) - ${configurationString}`} href="true" ref="downloadLink">Download <i className="fa fa-download ml-1" title="Download QRCode"></i></a>
+          </div>
+        </div>
+        <p className="mb-4 text-warning">Only the UID is unique, there might be more than one token with the same name and symbol.</p>
```
obiyankenobi: I thought we were gonna add this using a Bootstrap Alert, like the ones we have to say if a tx is valid or voided. It would have a yellow background and sit at the top of the screen, just like the valid/voided alert.

The way it is now looks a bit weird to me.

src/components/TxData.js
obiyankenobi: Actually, I think `tokens` might still need to be on the state, so we re-render the screen after we get the response from the API. But the rest of the suggestion is still valid.

src/components/TxData.js
pedroferreira1: Makes sense, added it.

src/screens/TokenDetail.js
pedroferreira1: When I coded I've tried both solutions and kept the one I thought was best. But to be honest both are good to me.

I've uploaded a screenshot with the bootstrap alert on the top.

Which one you think is better?

obiyankenobi: I prefer the new one, but I don't mind if we keep the other.

src/components/TxData.js
pedroferreira1: For the explorer the refactor makes total sense. Did it, thanks!

src/App.js
msbrogli: I agree. But, since you've added a MemoryStore here, isn't it better to just add it to the hathor-wallet-lib and use it from here? We have other repositories using MemoryStore as well.

src/screens/TokenDetail.js
msbrogli: Good. Thanks!

src/App.js
pedroferreira1: Makes sense. It's not for use in this explorer specifically, it's just in case another project wants a fast MemoryStore without need of implementing it.

Added it to the lib and I am just using here now.

PR #11, opened 2020-01-06
feat: UI improvements

pedroferreira1:
Changes:
- Disconnected peers are not shown anymore on Network screen;
- Removed some Statistics charts;
- Updated hashrate chart color to purple;
- Truncated hashrate number with prefix;
- Added search by address/tx_id on transactions screen.

Questions:
- Should we have a checkbox filter to show disconnected peers? Or even always show them but just in the end?
- Maybe add back one of the removed charts?
- With this new search, should we remove the search on the navigation (this one works only for tx_id)?
- The address search is exactly the same as the one in the wallet, when the user searches we show all results separating txs and blocks. Should we put together (without separation) and add a pagination?

image
image


public/index.html
```diff
@@ -6,7 +6,7 @@
     <meta name="theme-color" content="#000000">
     <link rel="manifest" href="%PUBLIC_URL%/manifest.json">
     <link rel="shortcut icon" href="%PUBLIC_URL%/favicon.ico">
-    <title>Hathor Network - Admin</title>
+    <title>Hathor Network - Explorer</title>
```
msbrogli: I would change to "Hathor Public Explorer", or just "Explorer - Hathor Network". What do you think?

src/screens/Dashboard.js
```diff
@@ -50,15 +51,13 @@ class Dashboard extends React.Component {
   }
 
   render() {
+    const blocks = this.props.data.length > 0 ? this.props.data[this.props.data.length - 1].blocks : '';
+    const transactions = this.props.data.length > 0 ? this.props.data[this.props.data.length - 1].transactions : '';
     return (
       <div className="content-wrapper">
-        <LineChartRealTime data={this.props.data} getX={this.getXData} getY={this.getYTxRate} unit="tx/s" title={["Tx Rate"]}/>
-        <LineChartRealTime data={this.props.data} getX={this.getXData} getY={this.getYTx} title={["Transactions"]} />
-        <LineChartRealTime data={this.props.data} getX={this.getXData} getY={this.getYBlock} title={["Blocks"]}/>
```
msbrogli: Both the number of blocks and the number of transactions are useful pieces of information. I think we should either keep the chart (over time) or show the latest number.

src/screens/DashboardTx.js
```diff
@@ -92,6 +141,8 @@ class DashboardTx extends React.Component {
 
     return (
       <div className="content-wrapper">
+        <p className="mt-4 mb-4">Here you can see the most recent transactions and blocks of the network.</p>
+        <SearchTx {...this.props} newData={this.newData} resetData={this.resetData} />
```
msbrogli: I think we should use the search text field at the top of the explorer. Then, after searching, we can show it (or not). What do you think?

src/screens/Dashboard.js
pedroferreira1: I've deleted the chart but added the information. Check https://github.com/HathorNetwork/hathor-explorer/pull/11/files#diff-fb9617d2e3aab8062b6397c120d67398R58 also see how it is in the screen shot

src/screens/DashboardTx.js
pedroferreira1: At first I thought it would be weird to have this search on top in the mobile but I will give it a try, probably is better for usability

public/index.html
pedroferreira1: I agree, changed to "Explorer - Hathor Network"

src/screens/DashboardTx.js
obiyankenobi: If we can have both on the address bar it's better for saving space on the main screen. But I'm also concerned about how it'll look on mobile.

src/components/Network.js
```diff
@@ -149,7 +149,7 @@ class Network extends React.Component {
         if (isConnected) {
           return renderConnected(peer, conn);
         } else {
-          return renderDisconnected(peer);
+          return null;
```
obiyankenobi: It seems useful if we still display peers that were recently disconnected, say less than 10 minutes ago.

pedroferreira1: We currently don't have this information in the API. I will open an issue and, if we find it relevant, we improve the API and add it here.

src/screens/Dashboard.js
msbrogli: Seems good!

msbrogli: Thanks!

PR #15, opened 2020-01-11
feat: URL history on transaction list browsing

pedrox:
This change adds URL history support (shareable URLs, back/forward browser support, etc.) to pages displaying transaction history.

**Affected Pages**

All pages displaying transaction history:
* [Blocks](https://explorer.hathor.network/blocks/)
* [Transactions](https://explorer.hathor.network/transactions/)
* [Token Detail](https://explorer.hathor.network/token_detail/00000eaee9f22c832b2b3affc71ce8ab2254957723557d0679dbbdc8a7fcdf4f)

**User Experience**

No UX change on the initial view of those pages. As the user navigates the transaction history using paging buttons Next/Previous, the URL parameters are added and updated accordingly. When the user gets back to the most recent page, the URL parameters are removed.

The introduced URL parameters maps 1:1 with the arguments of the [underlying API call](https://docs.hathor.network/#operation/transaction). Here's a sample path: `/blocks/?ts=1578703463&hash=0000000000040fa195bd7f01394bc4d4624488d5379cc0ee02e108d976d50933&page=next`

**Drawbacks**

~In this implementation, the Transaction component will update the URL assuming there are no URL query params other than its own. There are no URL query params in the three pages where this component is displayed.~

Also URLs are longer than they could be. This is actually a limitation of the back-end API. We could consider implementing a pagination API with a single continuation token and two mutually exclusive query params like `?after=<token>` or `?before=<token>`, where token could be composed by the timestamp and some least significant bytes of the hash. Although it's non-unique, it's short and very unlikely to collide. In case of collisions on page breaks, it's safe and fine to show duplicate records on two pages.

### DIFF
--- a/src/components/Transactions.js
+++ b/src/components/Transactions.js
@@ -7,6 +7,8 @@
 
 import React from 'react';
 import ReactLoading from 'react-loading';
+import { Link } from 'react-router-dom';
+import { isEqual } from 'lodash';
 import {TX_COUNT} from '../constants';
 import TxRow from './TxRow';
 import helpers from '../utils/helpers';
@@ -14,6 +16,22 @@ import WebSocketHandler from '../WebSocketHandler';
 import colors from '../index.scss';
 
 
+/**
+ * Displays transactions history in a table with pagination buttons. As the user navigates through the history,
+ * the URL parameters 'ts', 'hash' and 'page' are updated.
+ *
+ * Either all URL parameters are set or they are all missing.
+ *
+ * Example 1:
+ *   hash = "00000000001b328fafb336b4515bb9557733fe93cf685dfd0c77cae3131f3fff"
+ *   ts = "1579637190"
+ *   page = "previous"
+ *
+ * Example 2:
+ *   hash = "00000000001b328fafb336b4515bb9557733fe93cf685dfd0c77cae3131f3fff"
+ *   ts = "1579637190"
+ *   page = "next"
+ */
 class Transactions extends React.Component {
   constructor(props) {
     super(props);
@@ -27,13 +45,26 @@ class Transactions extends React.Component {
       loaded: false,
       hasAfter: false,
       hasBefore: false,
+      queryParams: this.obtainQueryParams(),
     }
   }
 
   componentDidMount() {
-    this.getData(true, null, null, '');
+    this.getData(this.state.queryParams);
 
     WebSocketHandler.on('network', this.handleWebsocket);
+  }
+
+  componentDidUpdate(prevProps, prevState) {
+    const queryParams = this.obtainQueryParams();
+
+    // Do we have new URL params?
+    if (!isEqual(this.state.queryParams, queryParams)) {
+      // Fetch new data, unless query params were cleared and we were already in the most recent page
+      if (queryParams.hash || this.state.hasBefore) {
+        this.getData(queryParams);
+      }
+    }
   }
 
   componentWillUnmount() {
@@ -65,7 +96,7 @@ class Transactions extends React.Component {
     }
   }
 
-  handleDataFetched = (data, first, page) => {
+  handleDataFetched = (data, queryParams) => {
     // Handle differently if is the first GET response we receive
     // page indicates if was clicked 'previous' or 'next'
     // Set first and last hash of the transactions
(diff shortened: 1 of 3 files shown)


src/components/Transactions.js
```diff
@@ -27,15 +28,38 @@ class Transactions extends React.Component {
...
+    if (JSON.stringify(this.state.queryParams) !== JSON.stringify(queryParams)) {
+      // Fetch new data, unless query params were cleared and we were already in the most recent page
+      if (queryParams.hash || this.state.hasBefore) {
+        this.getData(queryParams);
+      }
+      this.setState({ queryParams: queryParams });
+    }
+  }
+
+  obtainQueryParams() {
```
pedroferreira1: In this method here we assume all parameters we expect will be on the URL. If they are not there, the page will get stuck loading forever.

We could add an error handling like:

```
if (!(params.has('ts')) || !(params.has('hash')) || !(params.has('page'))) {
  window.history.replaceState({}, "", window.location.pathname); // The same URL param clean you've used
  return { timestamp: null, hash: null, page: null }
}
```

Something like that. What do you think?

src/components/Transactions.js
```diff
@@ -27,15 +28,38 @@ class Transactions extends React.Component {
...
+  componentDidUpdate(prevProps, prevState) {
+    const queryParams = this.obtainQueryParams();
+
+    // Do we have new URL params?
+    if (JSON.stringify(this.state.queryParams) !== JSON.stringify(queryParams)) {
+      // Fetch new data, unless query params were cleared and we were already in the most recent page
+      if (queryParams.hash || this.state.hasBefore) {
+        this.getData(queryParams);
+      }
+      this.setState({ queryParams: queryParams });
```
pedroferreira1: Just update it to `this.setState({ queryParams });` to keep the same code style, please.

src/components/Transactions.js
pedrox: The method has actually no such assumption. I could reproduce the "loading forever" issue you mentioned, by providing only `?hash=123`. However, this issue already exists and is more general: since the component currently [lacks error handling](https://github.com/HathorNetwork/hathor-explorer/blob/bb4e1998206835c64ac7957b7607d8e208b0b4f5/src/components/Transactions.js#L104-L107), any error on the initial API call would cause the same effect.

My understanding is that the above issue already exists and this change doesn't make it worse. If you agree, I propose we fix error handling in a separate PR.

Technical note: axiosInstance.js is not propagating rejections in its [`onRejected`](https://github.com/HathorNetwork/hathor-explorer/blob/bb4e1998206835c64ac7957b7607d8e208b0b4f5/src/api/axiosInstance.js#L22-L24) callback; it should `return Promise.reject(error)`.

On the other hand, errors on the subsequent API calls, although a bit less impactful, are made worse by this PR: since the URL gets updated upfront, there will be no effect clicking the same Next/Previous button again; the other button will force a new API call. Before this change, clicking the same button again would always trigger the API call. We should be able to fix that by updating `state.queryParams` only after we get a successful response, so that this state variable means "the latest queryParams we obtained results for". Thoughts?

Some context on `obtainQueryParams()`: it extracts the relevant URL params to a flat object format which is eventually stored in `state.queryParams`. It is used exactly twice:
1. On constructor to initialize state, which value is then used by `componentDidMount()` for the initial API call
2. On componentDidUpdate (also triggered on URL changes) to whether the params have been changed and if so, trigger a new API call.

pedroferreira1: I agree that we were not handling error right now but having the query params in the URL is easier for the user to send a bad request.

However I've checked the API and it's not properly handling the expected parameters there. I think before any error handling in the client side I should add some in the API.

I will create an issue in the full node to improve this error handling on the API and another one to add error handling here in the client. Is that ok?

pedrox: Sure. Agreed!

I've pushed a new version with the following changes:
* Address all comments so far
* Improved URL handling to preserve other parameters and fragment identifies (aka hash) if they exist
* Update `state.queryParams` only after getting successful response, to mitigate issues as described above. Now if a server error occurs, the user can simply retry by clicking the pagination button again. As discussed, proper error handling, like displaying an error message, should be implemented in a separate PR.

src/components/Transactions.js
```diff
@@ -27,15 +32,28 @@ class Transactions extends React.Component {
...
+    this.getData(this.state.queryParams);
 
     WebSocketHandler.on('network', this.handleWebsocket);
   }
 
+  componentDidUpdate(prevProps, prevState) {
+    const queryParams = this.obtainQueryParams();
+
+    // Do we have new URL params?
+    if (JSON.stringify(this.state.queryParams) !== JSON.stringify(queryParams)) {
```
marco-souza: You could use the mehtod `shouldComponentUpdate` to avoid rerendering if the url didn't change. 

> Use `shouldComponentUpdate()` to let React know if a component’s output is not affected by the current change in state or props.
> 
> https://reactjs.org/docs/react-component.html#shouldcomponentupdate

msbrogli: Even though `JSON.stringify` usually works, it may fail because of the order the keys are created. In this case, we can either add a new dependency (such as `loadsh`) or create our own shallow comparison method. What do you think?

src/components/Transactions.js
```diff
@@ -7,13 +7,18 @@
...
 import {TX_COUNT} from '../constants';
 import TxRow from './TxRow';
 import helpers from '../utils/helpers';
 import WebSocketHandler from '../WebSocketHandler';
 import colors from '../index.scss';
 
 
+/**
+ * Displays transactions history in a table with pagination buttons. As the user navigates through the history,
+ * the URL parameters 'ts', 'hash' and 'page' are updated.
```
msbrogli: Suggestion to add in the comments:

```
* Either all URL parameters are set or they are all missing.
* 
* Example 1:
*   hash = "00000000001b328fafb336b4515bb9557733fe93cf685dfd0c77cae3131f3fff"
*   ts = "1579637190"
*   page = "previous"
*
* Example 2:
*   hash = "00000000001b328fafb336b4515bb9557733fe93cf685dfd0c77cae3131f3fff"
*   ts = "1579637190"
*   page = "next"
```

src/components/Transactions.js
pedroferreira1: If we decide to do it we should take care because the URL change is not the only thing that have to trigger a render on the page. One other thing that might do it is a ws message with a new transaction for the page (we re-render it in case we are in the first page).

I am still not 100% sure if it's worth it to do it now, since the page has a good performance. Following react's docs:

> This method only exists as a performance optimization. Do not rely on it to “prevent” a rendering, as this can lead to bugs.

Of course it would be good but we just need to take care and think about all possible scenarios.

pedroferreira1: We already have `lodash` on `wallet-lib` so it wouldn't be a new dependency to be installed. We could add it to this project and use.

pedrox: Thanks for the comments, guys. I plan to leave it as is since it was never meant to prevent rendering. Although it could, I agree with pedroferreira1 that it shouldn't. This logic only decides whether or not to trigger a new backend fetch call by comparing the current URL params with the params of the last processed backend response.

PR #17, opened 2020-01-20
Improvements on UI using new APIs

### LINKED ISSUE #16: UI improvements to show more information
PR link: https://github.com/HathorNetwork/hathor-explorer/pull/17
Wallet lib PR link: https://github.com/HathorNetwork/hathor-wallet-lib/pull/42

- Tokens list;
- Add height to transaction detail;
- Nonce in tx detail shouldn't have `toString()`. It's not working for merged mining blocks;
- Add height to statistics
- Should we differ in tx detail the block from merged mining? 
  - Set the type as 'Merged Mined Block' and the normal as 'Block'? Yes
  - Show aux_pow data instead of the nonce? No

pedroferreira1:
Closes #16 

- Add block height on Transaction Detail screen;
- Remov `toString()` to nonce, so screen is not broken for merged mined blocks anymore;
- Now it shows 'Merged mined block' as type on Transaction Detail screen;
- Add height of the best chain on Statistics screen;
- Add tokens list screen with navigation link to it.

image
image

### DIFF
--- a/src/screens/Dashboard.js
+++ b/src/screens/Dashboard.js
@@ -54,9 +54,11 @@ class Dashboard extends React.Component {
     const blocks = this.props.data.length > 0 ? this.props.data[this.props.data.length - 1].blocks : '';
     const transactions = this.props.data.length > 0 ? this.props.data[this.props.data.length - 1].transactions : '';
     const peers = this.props.data.length > 0 ? this.props.data[this.props.data.length - 1].peers : '';
+    const height = this.props.data.length > 0 ? this.props.data[this.props.data.length - 1].best_block_height : '';
     return (
       <div className="content-wrapper">
         <p><strong>Blocks: </strong>{blocks}</p>
+        <p><strong>Height of the best chain: </strong>{height}</p>
         <p><strong>Transactions: </strong>{transactions}</p>
         <p><strong>Peers: </strong>{peers}</p>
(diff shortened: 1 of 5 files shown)


src/screens/Dashboard.js
```diff
@@ -54,9 +54,11 @@ class Dashboard extends React.Component {
     const blocks = this.props.data.length > 0 ? this.props.data[this.props.data.length - 1].blocks : '';
     const transactions = this.props.data.length > 0 ? this.props.data[this.props.data.length - 1].transactions : '';
     const peers = this.props.data.length > 0 ? this.props.data[this.props.data.length - 1].peers : '';
+    const height = this.props.data.length > 0 ? this.props.data[this.props.data.length - 1].best_block_height : '';
     return (
       <div className="content-wrapper">
         <p><strong>Blocks: </strong>{blocks}</p>
+        <p><strong>Height of the best chain: </strong>{height}</p>
```
msbrogli: Maybe we should rename it just to height. Or Latest block height. What do you think?

pedroferreira1: I just feel it might be confusing to some people to see "Blocks" and "Height" only and they are really different numbers, that's why I used "Height of the best chain"

I could use "Latest block height" but still the last block could be outside the best chain, so this name would be wrong.

What do you think?

msbrogli: I think they say "Latest block" thinking always in the best blockchain. But we can explicitly say "Height of the best chain". It seems good as well. Go ahead!

PR #21, opened 2020-01-30
feat: search for an address

pedroferreira1:
image
image


obiyankenobi:
The main doubt I have is whether we want to display transactions for ALL tokens. An address may have 10+ different tokens associated, so it'd be a very long list to show there.

pedroferreira1:
> The main doubt I have is whether we want to display transactions for ALL tokens. An address may have 10+ different tokens associated, so it'd be a very long list to show there.

The transactions list does not bother me to be honest, I feel like if we have the information we should show it and it will be paginated. We could add an easy filter on top in the future to show only HTR data but this would require changes on the APIs

obiyankenobi:
> > The main doubt I have is whether we want to display transactions for ALL tokens. An address may have 10+ different tokens associated, so it'd be a very long list to show there.
> 
> The transactions list does not bother me to be honest, I feel like if we have the information we should show it and it will be paginated. We could add an easy filter on top in the future to show only HTR data but this would require changes on the APIs

I didn't express myself well. The transaction list is fine, as it'll be paginated. My concern is with the summary for each token display above the transaction list, as there may be a lot of tokens.

pedroferreira1:
> > > The main doubt I have is whether we want to display transactions for ALL tokens. An address may have 10+ different tokens associated, so it'd be a very long list to show there.
> > 
> > 
> > The transactions list does not bother me to be honest, I feel like if we have the information we should show it and it will be paginated. We could add an easy filter on top in the future to show only HTR data but this would require changes on the APIs
> 
> I didn't express myself well. The transaction list is fine, as it'll be paginated. My concern is with the summary for each token display above the transaction list, as there may be a lot of tokens.

Yeah, this I agree we should discuss. Even though it's a possibility to have lots of different tokens in the same address it's not common, that's why I decided to do this UI (which is the best for me in 99.9% of the cases).

I believe we shouldn't hide any information from the user. If we decide to change this UI because of this problem I would suggest adding a filter, or aggregate custom tokens.

This last option I thought about implementing but didn't feel it was needed right now. It would have a summary for HTR and a summary of the rest (I still think it might not be good to sum all spent/received custom tokens).

That's why I think we should go with the current UI which would be ok even with 10 tokens (in my opinion) and we decide later what to do with more complex cases, maybe with a designer.

msbrogli:
I agree with @obiyankenobi that this UX/UI may get weird with many tokens. Maybe we should filter per token using a dropdown, where HTR would be the default token. Finally, as we will have a given token, we will be able to show whether it is receiving or sending tokens for each transaction.

I liked showing these pieces of information, but I think we should highlight the balance and show the "Total received" and the "Total sent" separately.

What do you think?

pedroferreira1:
I agree that a screen with many tokens would be bad. As I said in my comment before the best way to handle it that I thought was to add a filter, as you also suggested. I've implemented an option with the filter, even though the screen might be confusing, it solves this problem. Will explain why I think it's confusing, which was my biggest concern (as I said in my comment).

- Added number of tokens in the first summary.
- Added a select of tokens (only in case the address has more than one token)
- If HTR exists in this address, it will be selected by default. Otherwise we select the first token.

Start screen:
image

User changed token:
image

My concern is about the rest of data. The first summary and the list don't change when token is changed, this might confuse the user. To solve the many tokens problem I think this is the best solution and it's already implemented.

What do you think @msbrogli @obiyankenobi? Go with the filter, remove the filter and go back to the previous solution, or try a new approach?

obiyankenobi:
> The first summary and the list don't change when token is changed, this might confuse the user.

I think the tx list should change, to only show transactions that have that token.

I actually agree with @pedroferreira1 , that an address is unlikely to receive several tokens, so I'd be okay with that. However, I liked @msbrogli 's suggestion of having the total received and total sent in each tx of the list (or just the balance of the tx). In this case, if we don't have a filter by token, we'd have to show the balance for each token in the tx list.

I see some scenarios:
1) We **don't** have the token filter:
  1.1. We also don't show the balance for the transaction. This is the first designed Pedro implemented.
  1.2. We show the balance by token (or total sent/received). In that case, we might have to show multiple balances for each tx in the list, one for each token.
2) We add the token filter: would make it easier to show the balance for each tx, as we'd show only for the selected token.

I can imagine people might want to see all the txs of the address, including all tokens, so forcing to use the filter might not be the best. I think the best would be having a flexible filter, where you can add several (or all) tokens. I'm not sure if that's too complicated to implement as a first version.

msbrogli:
> > The first summary and the list don't change when token is changed, this might confuse the user.
> 
> I think the tx list should change, to only show transactions that have that token.
> 
> I actually agree with @pedroferreira1 , that an address is unlikely to receive several tokens, so I'd be okay with that. However, I liked @msbrogli 's suggestion of having the total received and total sent in each tx of the list (or just the balance of the tx). In this case, if we don't have a filter by token, we'd have to show the balance for each token in the tx list.
> 
> I see some scenarios:
> 
> 1. We **don't** have the token filter:
>    1.1. We also don't show the balance for the transaction. This is the first designed Pedro implemented.
>    1.2. We show the balance by token (or total sent/received). In that case, we might have to show multiple balances for each tx in the list, one for each token.
> 2. We add the token filter: would make it easier to show the balance for each tx, as we'd show only for the selected token.
> 
> I imagine people might want to see all the txs of the address, including all tokens, so forcing to use the filter might not be the best. I think the best would be having a flexible filter, where you can add several (or all) tokens. I'm not sure if that's too complicated to implement as a first version.

I think it is too complicated. Even though there may a case in which the user would like to see all transactions of an address, no matter the token, it seems very unusual to me.

I'd keep it simple and let the user select only one token, just like we to in the wallets.

I also agree that we should show only transactions of the selected token, including the number of transactions of that token.

I also think that we can remove the total number of transactions for all tokens. I think it makes this interface more confusing because it won't match with the transaction list when there are more than one token.

src/components/AddressSummary.js
```diff
@@ -0,0 +1,199 @@
...
+  /**
+   * balance {Object} Object with balance of each token of this address indexed by tokenUid {uid1: {'received', 'spent'}}
+   * quantity {number} Number of transactions in this address
+   * loading {boolean} If is waiting server response
+   * errorMessage {String} Message to be shown in case of an error
+   * selectedToken {String} UID of the selected token when address has many
+   */
+  state = {
+    balance: {},
+    quantity: null,
```
msbrogli: I'd rename it to `number_of_transactions`, or `n_transactions`, or `transactions_count`. What do you think?

src/components/AddressSummary.js
```diff
@@ -0,0 +1,199 @@
...
+        </div>
+      );
+    }
+
+    const loadBalanceInfo = () => {
+      const balance = this.state.balance[this.state.selectedToken];
+      return (
+        <div className="card bg-light mb-3">
+          <div className="card-body">
+            Token: {renderTokenData()}<br />
```
msbrogli: Missing i18n markers.

src/components/AddressSummary.js
```diff
@@ -0,0 +1,199 @@
...
+  render() {
+    if (Object.keys(this.state.balance).length === 0) {
+      return null;
+    }
+
+    const loadMainInfo = () => {
+      return (
+        <div className="card text-white bg-dark mb-3">
+          <div className="card-body">
+            Address: {this.props.address}<br />
```
msbrogli: Missing i18n markers.

src/components/AddressSummary.js
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    const renderTokenOptions = () => {
+      return Object.keys(this.state.balance).map((uid) => {
+        const tokenData = this.state.balance[uid];
+        return <option value={uid} key={uid}>{tokenData.name} ({tokenData.symbol})</option>;
+      });
+      return this.props.balance.map((token) => {
+        // Show in the select only the tokens not already selected plus the current selection
+        if (this.state.selectedTokens.find((selectedToken) => selectedToken.uid === token.uid) === undefined || token.uid === this.state.selected.uid) {
```
msbrogli: Empty if?

src/components/Navigation.js
```diff
@@ -82,14 +87,14 @@ class Navigation extends React.Component {
             </ul>
             <div className="navbar-right d-flex flex-row align-items-center navigation-search">
               <div className="d-flex flex-row align-items-center">
-                <input className="form-control mr-2" type="search" placeholder="Search tx" aria-label="Search" ref="txSearch" onKeyUp={this.handleKeyUp} />
+                <input className="form-control mr-2" type="search" placeholder="Search tx or address" aria-label="Search" ref="txSearch" onKeyUp={this.handleKeyUp} />
```
msbrogli: Missing i18n markers.

src/components/Navigation.js
```diff
@@ -82,14 +87,14 @@ class Navigation extends React.Component {
...
                 <i className="fa fa-search pointer" onClick={this.search}></i>
               </div>
               <Version />
             </div>
           </div>
         </nav>
-        <HathorAlert ref="alertError" text="Invalid hash format" type="danger" />
+        <HathorAlert ref="alertError" text="Invalid hash format or address" type="danger" />
```
msbrogli: Missing i18n markers.

src/screens/AddressDetail.js
```diff
@@ -0,0 +1,104 @@
...
+    errorMessage: '',
+  }
+
+  componentDidMount() {
+    // Expects address on URL
+    this.updateAddress(this.props.match.params.address);
+  }
+
+  componentDidUpdate(prevProps) {
+    if (prevProps.match.params.address !== this.props.match.params.address) {
```
msbrogli: Shouldn't we create a new `AddressDetail` for every search? It would allow us to skip this.

src/components/AddressHistory.js
```diff
@@ -0,0 +1,227 @@
...
+    queryParams: this.pagination.obtainQueryParams(),
+  }
+
+  componentDidMount() {
+    this.getData(this.state.queryParams);
+
+    WebSocketHandler.on('network', this.handleWebsocket);
+  }
+
+  componentDidUpdate(prevProps, prevState) {
```
msbrogli: Maybe we should render a new `AddressHistory` instance when the parameters are changed. It would prevent this method.

src/components/AddressHistory.js
```diff
@@ -0,0 +1,227 @@
...
+      if (response.success) {
+        this.handleDataFetched(response, queryParams);
+      }
+    });
+  }
+
+
+  render() {
+    if (this.state.transactions.length === 0) {
+      return <p>This address does not have any transactions yet.</p>
```
msbrogli: Missing i18n markers.

src/components/AddressHistory.js
```diff
@@ -0,0 +1,227 @@
...
+
+    const loadPagination = () => {
+      if (this.state.transactions.length === 0) {
+        return null;
+      } else {
+        return (
+          <nav aria-label="Tx pagination" className="d-flex justify-content-center">
+            <ul className="pagination">
+              <li ref="txPrevious" className={(!this.state.hasBefore || this.state.transactions.length === 0) ? "page-item mr-3 disabled" : "page-item mr-3"}>
+                <Link className="page-link" to={this.pagination.paginationUrl({hash: this.state.firstHash, page: 'previous'})}>Previous</Link>
```
msbrogli: Missing i18n markers.

src/components/AddressHistory.js
```diff
@@ -0,0 +1,227 @@
...
+      }
+    }
+
+    const loadTable = () => {
+      return (
+        <div className="table-responsive mt-5">
+          <table className="table table-striped" id="tx-table">
+            <thead>
+              <tr>
+                <th className="d-none d-lg-table-cell">Type</th>
```
msbrogli: Missing i18n markers.

obiyankenobi:
Ok, let's keep the filter, then. And I agree about the total number of transactions. We could either remove it from the first box or have 'Number of transactions (all tokens)'. Would also be useful to add the number of transactions from a specific token.

pedroferreira1:
Ok, so I will update the UI to have the filter on the tokens and on the history. I will also need to update the list API to consider a token filter.

src/components/AddressSummary.js
pedroferreira1: This variable does not exist anymore

src/components/AddressSummary.js
pedroferreira1: Removed it, thanks!

src/components/AddressSummary.js
pedroferreira1: We don't have i18n in the explorer, should we do something?

src/components/AddressSummary.js
pedroferreira1: We don't have i18n in the explorer, should we do something?

src/components/Navigation.js
pedroferreira1: We don't have i18n in the explorer, should we do something?

src/components/Navigation.js
pedroferreira1: We don't have i18n in the explorer, should we do something?

src/screens/AddressDetail.js
pedroferreira1: The problem is that the URL does not change, only the parameter, so it does not call the `componentDidMount` again. I can only catch the URL change with `componentDidUpdate`. Do you know a better way to do it?

src/components/AddressHistory.js
pedroferreira1: We don't have i18n in the explorer, should we do something?

src/components/AddressHistory.js
pedroferreira1: We don't have i18n in the explorer, should we do something?

src/components/AddressHistory.js
pedroferreira1: We don't have i18n in the explorer, should we do something?

src/components/AddressHistory.js
pedroferreira1: This method prevents executing unnecessary API requests, checking when it's really needed to do it. How do you think it should be done?

pedroferreira1:
@msbrogli @obiyankenobi At first I started this PR with the idea to create separate components (AddressSummary and AddressHistory), so they could be reused in the future alone, in the case a screen would need only the summary of an address, for instance.

At the beginning this was not a big problem because it was a simple screen, even though I had to do some repeated checks in both components, it was not a big deal. After we decided to add a token filter, the code became more complex and I believe some parts of the code became a little bit confusing because I had to exchange information between siblings components. I thought about doing a refactor to put everything in a single component, however the code was already advanced and I wanted a first feedback from you.

What do you think?

pedroferreira1:
Screenshots of the new screen:

image
image

src/components/AddressHistory.js
```diff
@@ -0,0 +1,352 @@
...
+   * Also validates if this tx is compatible with the list
+   *
+   * @param {Object} tx Data of a newly arrived tx
+   */
+  updateListWs = (tx) => {
+    // We only add new tx/blocks if it's the first page
+    if (!this.state.hasBefore) {
+      if (this.props.shouldUpdate(tx, true)) {
+        let transactions = this.state.transactions;
+        let hasAfter = (this.state.hasAfter || (transactions.length === TX_COUNT && !this.state.hasAfter))
```
obiyankenobi: Final `&& !this.state.hasAfter` seems unnecessary.

src/components/AddressHistory.js
```diff
@@ -0,0 +1,352 @@
...
+  getData = (queryParams) => {
+    hathorLib.walletApi.getSearchAddress(this.props.address, TX_COUNT, queryParams.hash, queryParams.page, queryParams.token, (response) => {
+      if (response.success) {
+        this.handleDataFetched(response, queryParams);
+      }
+    });
+  }
+
+  /**
+   * We get the final balance for the search address and selected token
```
obiyankenobi: Took me a while to understand this docstring. It's important to say it's _one tx's_ balance, not the address total balance. When I read "the final balance for the search address" I imagined the total balance.

src/components/AddressHistory.js
```diff
@@ -0,0 +1,352 @@
...
+   *
+   * @param {Object} tx Transaction data
+   *
+   * @return {Number} Final balance value (can be negative value, in case we spent more than received for the search address)
```
obiyankenobi: Also add "Final **tx** balance"

src/components/AddressSummary.js
```diff
@@ -0,0 +1,220 @@
...
+ * address: Address to show summary
+ * shouldUpdate: Function that receives a tx data and returns if summary should be updated
+ * pagination: Instance of pagination class that handles the URL parameters
+ */
+AddressSummary.propTypes = {
+  address: PropTypes.string.isRequired,
+  shouldUpdate: PropTypes.func.isRequired,
+  pagination: PropTypes.instanceOf(PaginationURL).isRequired,
+};
+
```
obiyankenobi: I'll comment here so it's easier to follow a thread discussion.

> At the beginning this was not a big problem because it was a simple screen, even though I had to do some repeated checks in both components, it was not a big deal. After we decided to add a token filter, the code became more complex and I believe some parts of the code became a little bit confusing because I had to exchange information between siblings components. I thought about doing a refactor to put everything in a single component, however the code was already advanced and I wanted a first feedback from you.

Actually, having 2 separate components make sense. The problem I see is having the logic in the components. It'd be better form them to be functional components that only display the data and leave the logic to the screen element, that makes all API requests and updates components' props. 

So the screen element requests the balance and history and also passes functions as props to handle changes in the components, like you already have with `onRowClicked` (and others). I'd just follow this pattern until we remove all logic from components. I think this could greatly simplify the code.

pedroferreira1: I agree that seeing the code now this might be the better approach but my idea was to make the two component 'plug and play'. If we needed to use them elsewhere we wouldn't need to execute the requests again, capture the ws message and everything else, we would just add the components there and everything would work. 

To do it, each component should be independent but I don't know if it makes sense to do it now, since it increases the complexity of the code. I think what I have right now is a mix of both approaches.

obiyankenobi: It was not clear to me which approach you think has higher code complexity. I still believe that what I suggested is much simpler. Right now, we have AddressHistory updating an element in AddressSummary, for example.

I understand your point about it being self contained, but I think the logic should not be in this component.

src/components/AddressHistory.js
msbrogli: You're right. Thanks!

src/components/AddressSummary.js
msbrogli: No, sorry.

src/components/Navigation.js
msbrogli: No. Sorry.

src/components/Navigation.js
msbrogli: No. Sorry.

src/screens/AddressDetail.js
msbrogli: When the user searches for a new address, we can change the instance of the object of `AddressDetail`. So, react won't match with the current screen and will render the new one. I've done it in other projects. Maybe we should talk so I can understand if it really applies here, but I think it does.

src/components/AddressHistory.js
msbrogli: No. Sorry.

src/components/AddressHistory.js
msbrogli: No. Sorry.

src/components/AddressHistory.js
```diff
@@ -0,0 +1,352 @@
...
+    }
+  }
+
+  /**
+   * Update component state when new list data arrives
+   *
+   * @param {Object} data Response data from the server
+   * @param {Object} queryParams URL parameters
+   */
+  handleDataFetched = (data, queryParams) => {
```
msbrogli: I'd say `handleFetchedData`.

src/components/AddressSummary.js
msbrogli: I agree with @obiyankenobi that the components should only display the results. All other things, including request management and business logic should be handled elsewhere.

It seems to require a refactor of many components/screens because we've been mixing these things all over the code in this project. Even so, maybe we should "do it right" and separate those things in the new components/screens.

src/screens/AddressDetail.js
pedroferreira1: I still don't get it how I should do it without this check. When I am already in the search page and I just search for a new address, the method `this.props.history.push` won't trigger a new `componentDidMount`. This check is just to handle this situation

src/components/AddressHistory.js
```diff
@@ -0,0 +1,352 @@
...
+import PropTypes from 'prop-types';
+import PaginationURL from '../utils/pagination';
+
+
+class AddressHistory extends React.Component {
+  /**
+   * transactions {Array} List of transactions on the list
+   * loading {boolean} If is waiting response of data request
+   * firstHash {String} First hash of the list, used in pagination
+   * lastHash {String} Last hash of the list, used in pagination
```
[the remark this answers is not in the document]
pedroferreira1: We needed them in the past for an old code. Removed them

src/components/AddressHistory.js
```diff
@@ -0,0 +1,352 @@
...
+   * hasAfter {boolean} If has more elements after the list page
+   * hasBefore {boolean} If has more elements before the list page
+   * queryParams {Object} Object with URL parameters data
+   * numberOfTransactions {Number} Total number of transactions of the list
```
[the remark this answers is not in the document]
pedroferreira1: No. `transactions.length` is only the list we currently display. This total number of transactions is shown on the address summary as the total for this address and the selected token.

src/components/AddressHistory.js
pedroferreira1: It does not make any difference but I agree, it's unnecessary. Removed it

src/components/AddressHistory.js
pedroferreira1: Makes sense. Updated the docstring

src/components/AddressSummary.js
pedroferreira1: I've updated the code and now the screen has all the logic.

Can you do the refactor again, please? @msbrogli @obiyankenobi

src/components/AddressHistory.js
pedroferreira1: Done

src/components/AddressHistory.js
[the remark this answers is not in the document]
obiyankenobi: Got it, thanks

src/screens/AddressDetail.js
```diff
@@ -0,0 +1,359 @@
...
+      hasBefore,
+      queryParams,
+      numberOfTransactions: data.total,
+    });
+  }
+
+  /**
+   * Request data from server and update state balance
+   */
+  getSummaryData() {
```
obiyankenobi: All other methods use es6 style functions. Weird that this works without `bind` 🤔

src/components/AddressHistory.js
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    const loadPagination = () => {
+      if (this.props.transactions.length === 0) {
+        return null;
+      } else {
+        return (
+          <nav aria-label="Tx pagination" className="d-flex justify-content-center">
+            <ul className="pagination">
+              <li ref="txPrevious" className={(!this.props.hasBefore || this.props.transactions.length === 0) ? "page-item mr-3 disabled" : "page-item mr-3"}>
```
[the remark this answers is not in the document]
pedroferreira1: True. Removed it

src/components/AddressHistory.js
```diff
@@ -0,0 +1,194 @@
...
+                <Link className="page-link" to={this.props.pagination.paginationUrl({hash: getFirstHash(), page: 'previous'})}>Previous</Link>
+              </li>
+              <li ref="txNext" className={(!this.props.hasAfter || this.props.transactions.length === 0) ? "page-item disabled" : "page-item"}>
```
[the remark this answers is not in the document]
pedroferreira1: True. Removed it

src/screens/AddressDetail.js
```diff
@@ -0,0 +1,359 @@
...
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+import AddressSummary from '../components/AddressSummary';
+import AddressHistory from '../components/AddressHistory';
+import { transaction, walletApi } from '@hathor/wallet-lib';
+import PaginationURL from '../utils/pagination';
+import hathorLib from '@hathor/wallet-lib';
```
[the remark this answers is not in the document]
pedroferreira1: Forgot to remove the old import. Thanks

src/screens/AddressDetail.js
pedroferreira1: Yes, it's weird. It was working fine. But changed it anyway.

src/utils/pagination.js
```diff
@@ -0,0 +1,61 @@
...
+    }
+    window.history.replaceState({}, '', url.href);
+  }
+
+  /**
+   * Set URL parameters
+   *
+   * data {Object} with key as param name and value as it's value to be set
+   */
+  paginationUrl(data) {
```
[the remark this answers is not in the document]
pedroferreira1: True. Changed to `setURLParameters`. What do you think?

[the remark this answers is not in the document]
obiyankenobi: Good

src/screens/AddressDetail.js
msbrogli: I got it. The point is that the react is matching you `AddressDetail` object, instead of creating a new one. You can force react to create a new one, which is safer because you ensure it will get nothing from previous instance. I think that one way to do it is to put the search address in the key of the `AddressDetail`.

src/components/AddressHistory.js
msbrogli: My bad.

src/components/AddressSummary.js
pedroferreira1: Since you two approved after the refactor, I assumed all is good. So resolving this

src/screens/AddressDetail.js
pedroferreira1: `AddressDetail` is created in the `NavigationRoute`. I don't think it's common in react to have a key there. I would rather follow what they have in the `react-router` repository (https://github.com/ReactTraining/react-router/blob/c865bc6b331eabd853641dcc7e0224a7dce76f3b/docs/guides/ComponentLifecycle.md#fetching-data) where they capture url parameter changes exactly as I do in the explorer.

What do you think?

msbrogli: Ok. I agree that you're following good practices.

PR #25, opened 2020-05-08
feat: adding economics data on dashboard

pedroferreira1:
image

### DIFF
--- a/src/api/economicsAxiosInstance.js
+++ b/src/api/economicsAxiosInstance.js
@@ -12,7 +12,7 @@ const requestClient = () => {
   const defaultOptions = {
     baseURL: ECONOMICS_API_URL,
     headers: {
-      'Content-Type': 'application/json',
+      'Content-Type': 'text/plain',
     },
   }
 
--- a/src/api/economicsAxiosInstance.js
+++ b/src/api/economicsAxiosInstance.js
@@ -1,0 +1,28 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { ECONOMICS_API_URL } from '../constants.js';
+const axios = require('axios');
+
+const requestClient = () => {
+  const defaultOptions = {
+    baseURL: ECONOMICS_API_URL,
+    headers: {
+      'Content-Type': 'application/json',
+    },
+  }
+
+  let instance = axios.create(defaultOptions);
+  instance.interceptors.response.use((response) => {
+    return response;
+  }, (error) => {
+    console.log("ERROR RESPONSE");
+  });
+  return instance;
+}
+
+export default requestClient();
--- a/src/screens/Dashboard.js
+++ b/src/screens/Dashboard.js
@@ -6,10 +6,12 @@
  */
 
 import React from 'react';
+import ReactLoading from 'react-loading';
 import { connect } from "react-redux";
-import LineChartRealTime from '../components/LineChartRealTime';
-import AreaChartRealTime from '../components/AreaChartRealTime';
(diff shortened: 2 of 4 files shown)


pedroferreira1:
In my opinion the solution is not 100% because the supply values (total and circulating) are not being updated on real time.

There are 2 possible solutions there I see (both have problems):

1. Get this information every time we have a height change (bad because we will have requests every time and if we have many explorer pages opened, we will reach 1M lambda requests easily).
2. Add economics information on websocket dashboard data, so it can be updated just like the rest of the data on this page. (bad because it's called every 5 seconds, so the full node will calculate the mined tokens every time. Also we would need to add a hardcoded `premine_released_tokens` value, just like the one in the lambda function.

What are your thoughts @msbrogli @obiyankenobi ? I think we can merge with no real time update but open an issue to handle it soon.

msbrogli:
As the other statistics are updated in real time, I think it is weird to have static values. Maybe we should update it every minute or something like this. What do you think?

src/screens/Dashboard.js
```diff
@@ -6,10 +6,12 @@
  */
 
 import React from 'react';
+import ReactLoading from 'react-loading';
 import { connect } from "react-redux";
-import LineChartRealTime from '../components/LineChartRealTime';
```
obiyankenobi, line 10: If we're removing the charts, we should do it properly. Maybe we should have a PR just for that, as there are quite a few changes:
1. remove the chart libs from `package.json`;
2. remove files that won't be used anymore, like `components/LineChartRealTime.js`;
3. refactor the redux state. We currently keep an array of the latest data received. I understand we do it for the charts, so there's no need to continue that;

src/screens/Dashboard.js
```diff
@@ -6,10 +6,12 @@
...
-import AreaChartRealTime from '../components/AreaChartRealTime';
 import colors from '../index.scss';
+import helpers from '../utils/helpers';
+import economicsApi from '../api/economics';
```
obiyankenobi, line 13: Starting a thread to discuss updating or not the economic metrics in realtime.

@pedroferreira1 actually the data updates every second, which I think we should change to 5 or 10 seconds.

I like the second suggestion. Maybe we should cache the mined tokens info on the server, as it only needs to be calculated when there's a new block. That would require a refactor on hathor core.

One other idea is just calling the economics api when the height of the best chain changes. That would keep this 100% realtime and prevent polling. But there would be a call to calculate the mined tokens everytime there's a new block, for each explorer that's open, so maybe caching it as I mentioned above would be better.

src/api/economicsAxiosInstance.js
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+import { ECONOMICS_API_URL } from '../constants.js';
+const axios = require('axios');
+
+const requestClient = () => {
+  const defaultOptions = {
+    baseURL: ECONOMICS_API_URL,
+    headers: {
+      'Content-Type': 'application/json',
```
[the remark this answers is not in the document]
pedroferreira1: Yes, you are right. Fixed it

src/screens/Dashboard.js
pedroferreira1: I will create an issue to do that in another PR. I guess it's better not to mix those things.

obiyankenobi:
Can we close this PR? I think we will not do it for the moment and we can open an issue to implement it in the future.

I'd say it's better, before making any changes in the explorer, refactoring the storage and removing the charts.

pedroferreira1:
As discussed, this PR will be closed because it still needs to be discussed how the economics data will be updated in real time.

Two issues were created to track the changes here:
- https://github.com/HathorNetwork/hathor-explorer/issues/26
- https://github.com/HathorNetwork/hathor-explorer/issues/29

PR #31, opened 2020-08-28
refactor: removing hash rate chart and its dependencies

pedroferreira1:


src/reducers/index.js
```diff
@@ -25,33 +23,16 @@ import { DASHBOARD_CHART_TIME } from '../constants';
...
+  data: null,
   isVersionAllowed: undefined,
 };
 
 const rootReducer = (state = initialState, action) => {
   switch (action.type) {
     case 'dashboard_update':
-      let data = state.data.slice(0);
       let newData = action.payload;
       newData['date'] = new Date(newData.time*1000);
```
obiyankenobi, line 34: I think we don't need this anymore, as it was used for the charts.

obiyankenobi (approved):

src/reducers/index.js
pedroferreira1: That's true, done.

msbrogli (approved):

PR #33, opened 2020-09-02
fix: hashrate units should be divided by 1000 and not 1024 because of new convention

pedroferreira1:


src/utils/helpers.js
```diff
@@ -135,10 +135,11 @@ const helpers = {
 
   /**
    * Divide a big value to be used with prefixes
-   * The value is divided by 1024 while it can
+   * The value is divided by 1000 while it can, which is the newest convention
+   * for K, M, ... prefixes. Binary prefixes are now Ki, Mi, ...
```
obiyankenobi, line 139: Is there a link explaining it?

obiyankenobi (approved):

src/utils/helpers.js
pedroferreira1: There is this, don't know if it's worth it to add to the docstring.

https://en.wikipedia.org/wiki/Kilobyte

> In December 1998, the IEC addressed such multiple usages and definitions by creating prefixes such as kibi, mebi, gibi, etc., to unambiguously denote powers of 1024.[10] Thus the kibibyte, symbol KiB, represents 2^10  bytes = 1024 bytes. These prefixes are now part of the International System of Quantities. The IEC further specified that the kilobyte should only be used to refer to 1000 bytes.

obiyankenobi: Ok, agreed no link is needed.

jansegre (approved):
Just a comment. This isn't because of any new convention. I don't think powers of 1024 have ever been used for measuring hashrate. Haha!

msbrogli (approved):

PR #37, opened 2020-09-04
feat: stop realtime update address balance

pedroferreira1:
There are 3 changes in this PR:

1. Stop real time update of address balance and show a warning message that there is a new transaction for the address.

image

2. Sometimes the address search request was being called twice when searching, depending on the response time of the first request.

3. The address search screen is composed of a Balance component and a History component. Each component has a request (one to get the balance of the address and another one to get the history - this last one with pagination).

Both requests take a long time depending on the number of transactions, in my tests I had them taking ~5s each. Given that both request are sent almost on the same time, the explorer axios timeout is 10s and the executes the second request only after the first one, most of the time I couldn't get the history because I received a timeout.

I customized our lib axios to increase the timeout right now because I think it's bad for this to happen to the explorer user, however I believe we should take a look why the requests are taking this long to reply.

I could've changed the code to request the history only after the balance arrived but this wouldn't solve the problem completely because soon we will have addresses with more txs and we will have the problem again. So I believe we should just increase the timeout for now while we improve the endpoints performance on the full node, then reduce the timeout again, what do you think?

### DIFF
--- a/src/api/customAxiosInstance.js
+++ b/src/api/customAxiosInstance.js
@@ -1,0 +1,28 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import hathorLib from '@hathor/wallet-lib';
+
+/**
+ * Create axios instance settings base URL and content type  
+ * Besides that, it captures error to show modal error and save in Redux
+ *
+ * @module Axios
+ */
+
+/**
+ * Create an axios instance to be used when sending requests
+ *
+ * @param {callback} resolve Callback to be stored and used in case of a retry after a fail
+ * @param {number} timeout Timeout in milliseconds for the request
+ */
+const createRequestInstance = (resolve, timeout) => {
+  const instance = hathorLib.axios.defaultCreateRequestInstance(resolve, 30000);
+  return instance;
+}
+
+export default createRequestInstance;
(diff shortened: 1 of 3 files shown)


obiyankenobi (approved):

jansegre (approved):

src/api/customAxiosInstance.js
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Create an axios instance to be used when sending requests
+ *
+ * @param {callback} resolve Callback to be stored and used in case of a retry after a fail
+ * @param {number} timeout Timeout in milliseconds for the request
+ */
+const createRequestInstance = (resolve, timeout) => {
+  // Will override lib axios instance increasing the default request timeout
+  const instance = hathorLib.axios.defaultCreateRequestInstance(resolve, 30000);
```
msbrogli, line 24: 30 seconds? Even though it seems too much, we can keep it if the request processing is not cancelled when the request times out.

msbrogli (approved):

src/screens/AddressDetail.js
```diff
@@ -330,6 +333,7 @@ class AddressDetail extends React.Component {
         } else {
           return (
             <div>
+              {this.state.warningRefreshPage && <div className="alert alert-warning" role="alert"> There is a new transaction for this address. Please refresh the page to see the newest data.</div>}
```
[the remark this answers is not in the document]
pedroferreira1: Done

src/api/customAxiosInstance.js
pedroferreira1: As discussed privately, we will keep 30s for now until our endpoints have better performance.

PR #42, opened 2021-03-23
feat: supporting status API for sync-v2

thr0wn:


thr0wn:
For sync-v2 status API, no timestamps are shown:
explorer

src/components/Network.js
```diff
@@ -67,6 +67,28 @@ class Network extends React.Component {
...
+    return { synced_percent, general_percent };
+  }
+
+  getSyncV2Progress(sync_v2) {
+    const synced_percent = Math.min(100, Math.abs(sync_v2.progress) * 100);
+    const general_percent = 100 - synced_percent;
+    return { synced_percent, general_percent };
+  }
+
+  getSyncProgress(sync_v1, sync_v2) {
```
msbrogli, line 88: Maybe this method should receive `conn` directly and do the logic here (instead of in the method `renderConnected`). What do you think?

src/components/Network.js
```diff
@@ -67,6 +67,28 @@ class Network extends React.Component {
...
+    const synced_percent = 100 * (sync_v1.synced_timestamp - first_timestamp) / delta;
+    let general_percent = (100 * (sync_v1.latest_timestamp - first_timestamp) / delta) - synced_percent;
+    if (general_percent > 100) {
+      general_percent = 100;
+    }
+    return { synced_percent, general_percent };
+  }
+
+  getSyncV2Progress(sync_v2) {
+    const synced_percent = Math.min(100, Math.abs(sync_v2.progress) * 100);
```
msbrogli, line 83: Protection against `sync_v2.progress` out of the interval [0, 1]?

src/components/Network.js
```diff
@@ -103,15 +125,9 @@ class Network extends React.Component {
...
-        const latest_timestamp = this.state.dag.latest_timestamp;
-        const delta = latest_timestamp - first_timestamp;
-        const synced_percent = 100 * (sync.synced_timestamp - first_timestamp) / delta;
-        let general_percent = (100 * (sync.latest_timestamp - first_timestamp) / delta) - synced_percent;
-        if (general_percent > 100) {
-          general_percent = 100;
-        }
+        const sync_v1 = conn.plugins["node-sync-timestamp"];
+        const sync_v2 = conn.sync;
+        const { synced_percent, general_percent } = this.getSyncProgress(sync_v1, sync_v2);
```
msbrogli, line 130: You can add a `sync_version` variable that will be used later. What do you think?

src/components/Network.js
```diff
@@ -129,10 +145,12 @@ class Network extends React.Component {
             </div>
             <ul className="list-group list-group-flush">
               <li className="list-group-item">
-                <div>
-                  Synced timestamp: {dateFormatter.timestampToString(sync.synced_timestamp)}<br />
-                  Latest timestamp: {dateFormatter.timestampToString(sync.latest_timestamp)}
-                </div>
+                {sync_v1 ? (
+                  <div>
+                    Synced timestamp: {dateFormatter.timestampToString(sync_v1.synced_timestamp)}<br />
```
msbrogli, line 150: I'd like to have at least the latest timestamp in sync-v2. Is it possible? Maybe we need to change the design.

msbrogli (changes requested):

src/components/Network.js
```diff
@@ -129,10 +145,12 @@ class Network extends React.Component {
             </div>
             <ul className="list-group list-group-flush">
               <li className="list-group-item">
-                <div>
-                  Synced timestamp: {dateFormatter.timestampToString(sync.synced_timestamp)}<br />
-                  Latest timestamp: {dateFormatter.timestampToString(sync.latest_timestamp)}
-                </div>
+                {sync_v1 ? (
```
pedroferreira1, line 148: I thought about adding the `state` information here. Not sure if it's important but I guess it would be good just to know the status of the sync.

As described in the design: 
`state: can be unknown, behind, in-sync, ahead, to indicate where a peer state is in relation to us`

src/components/Network.js
jansegre: It's possible when we have the block of the remote peer, which is almost always the case, but not always, but it is technically possible to still have it.

But I'd rather make this something specific to sync-v2 rather than making it generic into the `sync` information, which doesn't have a design yet. I also don't quite understand the practical advantage of showing the timestamp instead of the height+hash (hich is what the sync-v2 will be able to show), timestamp is indirectly accessible.

In the odd case where a node might be mining in the future we wouldn't know the timestamp either, which is the only case I can think of where actually having the timestamp would be useful.

src/components/Network.js
```diff
@@ -103,15 +125,9 @@ class Network extends React.Component {
...
-        const first_timestamp = this.state.dag.first_timestamp;
-        const latest_timestamp = this.state.dag.latest_timestamp;
-        const delta = latest_timestamp - first_timestamp;
-        const synced_percent = 100 * (sync.synced_timestamp - first_timestamp) / delta;
-        let general_percent = (100 * (sync.latest_timestamp - first_timestamp) / delta) - synced_percent;
-        if (general_percent > 100) {
-          general_percent = 100;
-        }
+        const sync_v1 = conn.plugins["node-sync-timestamp"];
+        const sync_v2 = conn.sync;
```
jansegre, line 129: This isn't quite right, `conn.sync` is actually the common denominator of all sync protocols and not exclusive to `sync_v2`. It might make sense that we have a `conn.sync.version`, but rather than that I think it'd be best if the code was actually less dependent on the sync version, the way it is now looks more dependent.

Maybe if we make an object that has all the sync data where fields from sync-v1 are OK to be null, like:

```
{
  "progress": 0.5,
  "state": "behind",
  "latest_timestamp": null,
  "synced_timestamp": null
}
```

The only way `conn.sync` is undefined is when the node is outdated, which I don't think we have to cover (updated explorer only works on updated node is a reasonable compromise).

src/components/Network.js
thr0wn: I agree, what do you prefer?
image

image

src/components/Network.js
thr0wn: Agreed. Done.

src/components/Network.js
thr0wn: Protection added (`Math.min(100, Math.max(0, sync_v2.progress * 100))`).

src/components/Network.js
thr0wn: It will not be required if we adopt the new `sync.data` suggested [here](https://github.com/HathorNetwork/hathor-explorer/pull/42#discussion_r600788906).

src/components/Network.js
thr0wn: This new `sync data` seems a good idea, therefore I will make changes to expect this structure on the explorer.

src/components/Network.js
pedroferreira1: I think we shouldn't show 'in-sync'. We should have a pretty format like 'In sync'

thr0wn: Done

src/components/Network.js
thr0wn: Done

src/components/Network.js
thr0wn: I followed this `sync data` structure, thanks for your suggestion.

src/components/Network.js
thr0wn: Now I conditionally display `synced timestamp` and `latest timestamp`, so unless you want to not display timestamps, this requested change is resolved.

src/components/Network.js
```diff
@@ -11,6 +11,11 @@ import ReactLoading from 'react-loading';
 import dateFormatter from '../utils/date';
 import colors from '../index.scss';
 
+const SyncStatus = {
+  'in-sync': 'In sync',
```
msbrogli, line 15: I think we should use more descriptive texts here. For instance: "Synchronized", "This peer is behind us", and "This peer is ahead of us". But, in the public explorer, maybe we should use only "Synchronized" and "Not synchronized" with a question mark next to the "Not synchronized" with either "behind of us" ou "ahead of us".

src/components/Network.js
```diff
@@ -103,15 +132,12 @@ class Network extends React.Component {
...
-        const sync = conn.plugins["node-sync-timestamp"];
-        const first_timestamp = this.state.dag.first_timestamp;
-        const latest_timestamp = this.state.dag.latest_timestamp;
-        const delta = latest_timestamp - first_timestamp;
-        const synced_percent = 100 * (sync.synced_timestamp - first_timestamp) / delta;
-        let general_percent = (100 * (sync.latest_timestamp - first_timestamp) / delta) - synced_percent;
-        if (general_percent > 100) {
-          general_percent = 100;
-        }
+        const sync_data = {
```
msbrogli, line 135: What fields should we expect in this object? Add them and their types in the comments, please.

msbrogli (approved):

src/components/Network.js
thr0wn: Done, thanks.

src/components/Network.js
```diff
@@ -67,6 +72,30 @@ class Network extends React.Component {
...
+
+  calcSyncProgressFromFraction(sync_data) {
+    const { progress = 0 } = sync_data;
+    const synced_percent = Math.min(100, Math.max(0, progress * 100));
+    const general_percent = 100 - synced_percent;
+    return { synced_percent, general_percent };
+  }
+
+  getSyncProgress(sync_data) {
+    return sync_data.synced_timestamp && sync_data.latest_timestamp ?
```
pedroferreira1, line 95: We want to always use the new data when it's available, so I would check if we have the new API key `sync` and use the timestamps only if it's no available.

pedroferreira1 (changes requested):

src/components/Network.js
thr0wn: You're right. Now it will only use timestamps to build a polyfill of the new sync data structure ([here](https://github.com/HathorNetwork/hathor-explorer/blob/0b8efbf7ba0dafb1e93da0a8b0d888c20ac0f885/src/components/Network.js#L153)).

pedroferreira1 (approved):

src/components/Network.js
```diff
@@ -11,6 +11,19 @@ import ReactLoading from 'react-loading';
...
+const SyncStates = {
+  IN_SYNC: 'in-sync',
+  BEHIND: 'behind',
+  AHEAD: 'ahead',
+  UNKNOWN: 'unknown',
+};
+
+const SyncStatesDescription = {
+  [SyncStates.IN_SYNC]: 'Synchronized',
+  [SyncStates.BEHIND]: <span title="Behind of us">Not synchronized <i class="fa fa-question-circle"></i></span>,
```
msbrogli, line 23: I'd change it to "Synchronizing..." for both of them.

thr0wn: Done

jansegre (approved):

Issue #45, opened 2021-04-30
Improve UI when we are unable to decode script


pedroferreira1:
We should show the output script in base64 when we are unable to decode it.

The output unable to decode should also respect the `token_data` information to show the correct token.

Testnet tx with unable to decode: https://explorer.testnet.hathor.network/transaction/00000573e9ff3f594176f47c05457f1b3474a9fd6d79d5c5885f0f2106d800c9

PR #47, opened 2021-04-30
feat(tx): Show script data when unable to decode

giovanecosta:
https://github.com/HathorNetwork/hathor-explorer/issues/45

Outputs with unknown script now will look like this: 
image

### DIFF
--- a/src/components/TxData.js
+++ b/src/components/TxData.js
@@ -178,7 +178,7 @@ class TxData extends React.Component {
 
     const renderOutputToken = (output) => {
       return (
-        <strong>{this.getOutputToken(hathorLib.wallet.getTokenIndex(output.decoded.token_data))}</strong>
+        <strong>{this.getOutputToken(hathorLib.wallet.getTokenIndex(output.token_data))}</strong>
       );
     }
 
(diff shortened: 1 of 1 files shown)


src/components/TxData.js
```diff
@@ -215,15 +215,20 @@ class TxData extends React.Component {
...
-          return renderNanoContractMatchValues(decoded);
+          return renderNanoContractMatchValues(output.decoded);
         default:
-          return 'Unable to decode';
+          let script = output.script;
+          try {
+            script = atob(output.script)
+          } catch {}
+
+          return `Unable to decode script: ${script.trim()}`;
```
msbrogli, line 231: I feel we can't just do this because of injection attack vectors.

msbrogli (changes requested):

src/components/TxData.js
giovanecosta: @msbrogli It's already safe by design. If i want to put some HTML dynamically here, i have to do it explicitly.

msbrogli: Ok. Great. Thanks!

msbrogli (changes requested):
I'm good to approve but your commit messages are not following the Semantic Commits.

pedroferreira1 (approved):

msbrogli (approved):

Issue #46, opened 2021-04-30
Mark genesis transactions to avoid confusion


pedroferreira1:
We have 3 genesis transactions (one block and two transactions), that are custom transactions without parents.

We must mark those transactions as genesis in the explorer to avoid confusion when seeing them.

```
Testnet: [0000033139d08176d1051fb3a272c3610457f0c7f686afbe0afe3d37f966db85, 00e161a6b0bee1781ea9300680913fb76fd0fac4acab527cd9626cc1514abdc9, 00975897028ceb037307327c953f5e7ad4d3f42402d71bd3d11ecb63ac39f01a]
Mainnet: [000006cb93385b8b87a545a1cbb6197e6caff600c12cc12fc54250d39c8088fc, 0002d4d2a15def7604688e1878ab681142a7b155cbe52a6b4e031250ae96db0a, 0002ad8d1519daaddc8e1a37b14aac0b045129c01832281fb1c02d873c7abbf9]
```

PR #48, opened 2021-04-30
feat(tx): mark genesis blocks and transactions

giovanecosta:
https://github.com/HathorNetwork/hathor-explorer/issues/46

Genesis block/transactions pages now will look like this:
image
image


src/components/TxData.js
```diff
@@ -431,14 +431,21 @@ class TxData extends React.Component {
       );
     }
 
+    const renderGenesisMark = (hash) => {
+      console.log(hash)
```
pedroferreira1, line 435: Debug log

src/constants.js
```diff
@@ -25,12 +25,18 @@ export const DASHBOARD_BLOCKS_COUNT = 6;
...
 
 export const GENESIS_BLOCK = [
-  '000000a0f82cfee5431e03b071364970861ffa1b0633f73ca7f462987ec34195'
+  '000000a0f82cfee5431e03b071364970861ffa1b0633f73ca7f462987ec34195', // Default
+  '0000033139d08176d1051fb3a272c3610457f0c7f686afbe0afe3d37f966db85', // Testnet
+  '000006cb93385b8b87a545a1cbb6197e6caff600c12cc12fc54250d39c8088fc' // Mainnet
 ]
 
 export const GENESIS_TX = [
   '000000831cff82fa730cbdf8640fae6c130aab1681336e2f8574e314a5533849',
```
pedroferreira1, line 34: I think we can remove the first two transactions. It's an old genesis test

src/constants.js
```diff
@@ -25,12 +25,18 @@ export const DASHBOARD_BLOCKS_COUNT = 6;
 export const DASHBOARD_TX_COUNT = 6;
 
 export const GENESIS_BLOCK = [
-  '000000a0f82cfee5431e03b071364970861ffa1b0633f73ca7f462987ec34195'
```
pedroferreira1, line 28: I think we can remove the first hash. It's an old genesis test

pedroferreira1 (approved):

src/constants.js
```diff
@@ -25,12 +25,18 @@ export const DASHBOARD_BLOCKS_COUNT = 6;
 export const DASHBOARD_TX_COUNT = 6;
 
 export const GENESIS_BLOCK = [
```
msbrogli, line 27: I don't think we should mix hashes from mainnet and testnet. Maybe we should have two config files: one for the mainnet and another for testnet. @giovanecosta @pedroferreira1 What do you think?

msbrogli (changes requested):

src/constants.js
pedroferreira1: Maybe we can start with two different variables in the same constants file? The lib works like this nowadays and we still need to implement this issue (https://github.com/HathorNetwork/hathor-wallet-lib/issues/161) which discusses exactly having a configuration file and being able to easily customize it.

To be honest, the best approach would be for the explorer to get this information (genesis hashes) from the lib in the future but I feel we can start with this solution for now.

msbrogli: Ok. We can split into two different variables (one for the testnet and another one for the mainnet).

msbrogli (changes requested):
You need to sign your commits and fix the commit messages.

src/constants.js
giovanecosta: I was looking into the code trying to understand if we could use environment variables to do it, so, we can keep one constant and the explorer agnostic from which network it is connecting. But i realized that we deploy it on Cloudfront, so, no env vars :sweat_smile:

In this case, i agree with two different variables and i'll make a function into `utils.helpers` that checks the network and return if the tx is genesis or not

pedroferreira1 (approved):

src/constants.js
msbrogli: The site is static so there should be a way to inject code based on envvars in compile time.

src/utils/helpers.js
```diff
@@ -156,6 +157,41 @@ const helpers = {
...
+   * Checks if transaction is a Genesis transaction
+   *
+   * @param {string} hash Hash of transaction
+   *
+   * @return {boolean} true if is genesis, false otherwise
+   * @memberof Helpers
+   * @inner
+   */
+  isGenesisTx(hash) {
+    if (hathorLib.network.getNetwork().name === 'mainnet') {
```
msbrogli, line 171: Move it to a separate method `isMainnet()`.

msbrogli (changes requested):

msbrogli (approved):

Issue #43, opened 2021-04-07
[Design] Hathor Explorer v1


giovanecosta:
- Feature Name: hathor_explorer_v1
- Start Date: 2021-04-09
- RFC PR: 
- Hathor Issue: 
- Author: Giovane Tomaz <gigio.coder@gmail.com>

# Summary
[summary]: #summary

Version 1 of Hathor Explorer connecting to a exclusive service (instead of connect to the full-node directly) with some new features.

# Motivation
[motivation]: #motivation

Hathor Explorer currently accesses the `full-node` directly increasing load, specially with some requests that leads to expensive operations. There are a lot of information that we can cache, store in an simpler way that is easy to retrieve or even request from other running services. Therefore we'll access the `full-node` only when is strictly necessary, preferably for simple operations.

# Guide-level explanation
[guide-level-explanation]: #guide-level-explanation

As the main goal of this upgrade is to provide same Hathor Network data as usual but drastically reducing the overload that some operations can cause on `full-node`, we are creating a new project called `hathor-explorer-service` (maybe `hathor-explorer-api` is a better name?).

At first, this new project will be responsible for:
- Store data that is relevant to the explorer
- Handle WS connections that was handled by the `full-node`
- Handle Web Requests that was handled by the `full-node`
- Listen to updates from `full-node` and store data in its own database
- Make requests for other services that are already detached from `full-node` 
- Cache most of these requests tha still need to be made to `full-node` 

Deploy will be done by Serverless framework (lambdas and API Gateway) and Terraform (all other things)

The following image illustrates how the parts of the services will interact:
![image](https://user-images.githubusercontent.com/698586/119893068-4fda7c00-bf11-11eb-83b2-c658a50cea57.png)

**Note:* `hathor-running-services` is an abstraction for services that are already currently running.

***Note:* `hathor-explorer-lib` might be extracted from `hathor-explorer` in the future.

## Hathor Explorer Service
[guide-level-explanation/hathor-explorer-service]: #guide-level-explanation/hathor-explorer-service

Hathor explorer service will be responsible for provide data to be consumed by explorer. Some of this data will be obtained by requesting other services and some will be stored on service own database.

### Aurora MySQL
[guide-level-explanation/hathor-explorer-service/aurora-mysql]: #guide-level-explanation/hathor-explorer-service/aurora-mysql

Aurora MySQL will not be deployed by Serverless framework. 

[...]

PR #52, opened 2021-05-10
Design: Hathor explorer v1

giovanecosta:
https://github.com/HathorNetwork/hathor-explorer/issues/43


hathor_explorer_v1.md
```diff
@@ -0,0 +1,772 @@
...
+[guide-level-explanation/hathor-explorer-service]: #guide-level-explanation/hathor-explorer-service
+
+Hathor explorer service will be responsible for provide data to be consumed by explorer. Some of this data will be obtained by requesting other services and some will be stored on service own database.
+
+### Postgresql
+[guide-level-explanation/hathor-explorer-service/postgresql]: #guide-level-explanation/hathor-explorer-service/postgresql
+
+Postgresql will run behind a RDS proxy to handle lambda connections and will not be deployed by Serverless framework. 
+
+We choose Postgresql because we want to have a nice and complex features in the feature and a good relational database is essential to give us enough flexibility to do it.
```
luislhl, line 50: I found this sentence a little bit odd. For me, a non-relational database is the one to go for maximum flexibility. Relational databases have less flexibility, because they have more mechanisms to assure consistency.

I think the main factor to choose between relational and non-relational should be the data we will have to work with. If the data is relational in its nature or needs to be highly structured, a relational database would be better.
Makes sense?

Probably we would have to model the data that we expect to deal with in the future to have a better understanding on this, unless you already have a good idea.

hathor_explorer_v1.md
```diff
@@ -0,0 +1,766 @@
...
+It's a good idea to separate things inside the project first and extract to a different one later. So it will still be part of [`hathor-explorer`](#guide-level-explanation/hathor-explorer) for now.
+
+## Hathor Explorer
+[guide-level-explanation/hathor-explorer]: #guide-level-explanation/hathor-explorer
+
+The existing project will not have any big visual modification except by network visualization, token list pagination and the hashrate x time graph that will be "all time" instead of starting only by the time the page is loaded.
+
+Furthermore business logic will be extracted from React components and API calls will be modified, if needed, to fit the new responses from new [guide-level-explanation/hathor-explorer-service](#guide-level-explanation/hathor-explorer-service)
+
+> **Estimated total monthly cost:** $ 56.00 with a huge margin of accesses
```
luislhl, line 182: What's this cost based on? What kinds of AWS services are considered in this?

giovanecosta: @luislhl each section above have the estimated cost. This is the sum of all. I'm using AWS "pricing" pages as reference.

luislhl (approved):

PR #53, opened 2021-05-12
feat(tx): app load screen

giovanecosta:
## When some error occur with API:

https://user-images.githubusercontent.com/698586/118285552-f487a300-b4a7-11eb-9c71-c57dded10a8f.mov

## Expected behavior (slow 3G emulated):

https://user-images.githubusercontent.com/698586/118285574-fbaeb100-b4a7-11eb-9dc0-2139bb8c635f.mov

## Changing pages while loading:

https://user-images.githubusercontent.com/698586/118285604-036e5580-b4a8-11eb-8eb1-75898511d941.mov

## Taking more than 3 seconds to load:

https://user-images.githubusercontent.com/698586/119036241-ea254780-b986-11eb-8534-cbe4d1ee4134.mp4


src/App.js
```diff
@@ -80,7 +82,13 @@ class Root extends React.Component {
   render() {
     if (this.props.isVersionAllowed === undefined) {
       // Waiting for version
-      return null;
+      return (
+        <Router>
+          <>
+            <Navigation />
```
andreabadesso, line 88: What happens if the user clicks the navigation buttons while the app is loading?

giovanecosta: The page just change, but the loading component remains. I've added a video showing it on PR description

src/App.js
```diff
@@ -80,7 +85,18 @@ class Root extends React.Component {
   render() {
     if (this.props.isVersionAllowed === undefined) {
       // Waiting for version
-      return null;
+      return (
+        <Router>
+          <>
```
pedroferreira1, line 90: Why do you need this? `<>`

src/actions/index.js
```diff
@@ -7,4 +7,6 @@
 
 export const dashboardUpdate = data => ({ type: "dashboard_update", payload: data });
 
-export const isVersionAllowedUpdate = data => ({ type: "is_version_allowed_update", payload: data });
\ No newline at end of file
+export const isVersionAllowedUpdate = data => ({ type: "is_version_allowed_update", payload: data });
+
+export const apiLoadErrorUpdate = data => ({ type: 'api_load_error_update', payload: data })
```
pedroferreira1, line 12: Missing `;` in end of the line. (Maybe we should create an issue to add a linter in the near future)

src/components/Loading.js
```diff
@@ -0,0 +1,28 @@
...
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+
+import ReactLoading from 'react-loading';
+import colors from '../index.scss';
+
+
+const Loading = (props) => {
```
pedroferreira1, line 14: I think we've been using 2 spaces of indentation in js. Maybe we should keep it in all files, what do you think? The linter would help solving this kind of problems?

src/App.js
```diff
@@ -53,13 +56,15 @@ class Root extends React.Component {
...
     hathorLib.axios.registerNewCreateRequestInstance(createRequestInstance);
+    this.props.apiLoadErrorUpdate({apiLoadError: false});
 
     versionApi.getVersion().then((data) => {
       hathorLib.network.setNetwork(data.network.split('-')[0]);
       this.props.isVersionAllowedUpdate({allowed: helpers.isVersionAllowed(data.version)});
     }, (e) => {
       // Error in request
       console.log(e);
+      this.props.apiLoadErrorUpdate({apiLoadError: true});
```
pedroferreira1, line 67: I like this but I understand that this only works if the first request fails (the version request), right? This is the home screen and it already help us in the case the screen becomes white when the explorer API is not working. I think we should keep this PR only with this but create an issue to add this `apiLoadError: true` redux value when other APIs fail as well.

pedroferreira1 (dismissed):

src/App.js
giovanecosta: This is a syntax sugar for `React.fragment`. So, we can have just one child of `Router` without get html dirty with a meaningless `div`. 
And we need an element there because: `A <Router> may have only one child element`.

Ref: https://reactjs.org/docs/fragments.html

andreabadesso (dismissed):

src/App.js
giovanecosta: Sure. I agree. I just think we have to discuss how we'll show it on page, because it can be a case of service shutdown, a momentary problem, just that specific endpoint not working or a 402 error, for example.

andreabadesso (dismissed):

src/App.js
msbrogli: I agree that we should work on it in a separate PR. So, we should at least create a new issue for that.

src/App.js
```diff
@@ -80,7 +85,18 @@ class Root extends React.Component {
...
     if (this.props.isVersionAllowed === undefined) {
       // Waiting for version
-      return null;
+      return (
+        <Router>
+          <>
+            <Navigation />
+            { this.props.apiLoadError ? 
+              <div className="content-wrapper">
+                <h3 className="text-danger">Error loading API. Please try again later.</h3>
```
msbrogli, line 94: I'd say "Error loading the explorer. Please reload the page to try again.".

src/components/Loading.js
```diff
@@ -0,0 +1,28 @@
...
+  return (
+    <div className="loading-wrapper">
+      <ReactLoading {...props} />
```
msbrogli, line 17: If it takes longer than 3 seconds, should we write "It is still loading... Please, be patient."?

msbrogli (dismissed):

src/App.js
giovanecosta: Issue created: https://github.com/HathorNetwork/hathor-explorer/issues/57

src/components/Loading.js
```diff
@@ -0,0 +1,39 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React, { useEffect, useState } from 'react';
```
pedroferreira1, line 8: I like this but my first impression was that our React version did not support this (https://reactjs.org/docs/hooks-reference.html). Support was added on 16.8.0 and our package.json uses 16.5.0.

However I was looking at package-lock.json and, for some reason, React version is 16.12.0 the since Feb 2020. I think it's safe to assume v16.12.0 works, just think we should upgrade the version in package.json as well

pedroferreira1 (approved):

andreabadesso (approved):

andreabadesso (approved):

pedroferreira1 (approved):

PR #54, opened 2021-05-19
feat(tx): show media for special tokens

giovanecosta:
# No media:
image

# Image:
image

# Video:
image

# Melt/Mint info hover:
image

### DIFF
--- a/src/api/axiosInstance.js
+++ b/src/api/axiosInstance.js
@@ -5,8 +5,12 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import { BASE_URL } from '../constants.js';
+import { BASE_URL, BASE_URL_V1 } from '../constants.js';
 const axios = require('axios');
+
+const errorHandler = (error) => {
+  console.log("ERROR RESPONSE", error);
+}
 
 const requestClient = () => {
   const defaultOptions = {
@@ -19,10 +23,25 @@ const requestClient = () => {
   let instance = axios.create(defaultOptions);
   instance.interceptors.response.use((response) => {
     return response;
-  }, (error) => {
-    console.log("ERROR RESPONSE");
-  });
+  }, errorHandler);
+  return instance;
+}
+
+const requestClientV1 = () => {
+  const defaultOptions = {
+    baseURL: BASE_URL_V1,
+    headers: {
+      'Content-Type': 'application/json',
+    },
+  }
+
+  let instance = axios.create(defaultOptions);
+  instance.interceptors.response.use((response) => {
+    return response;
+  }, errorHandler);
   return instance;
 }
 
 export default requestClient();
+
+export { requestClientV1 };
--- a/src/api/axiosInstance.js
+++ b/src/api/axiosInstance.js
@@ -5,7 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import { BASE_URL, BASE_URL_V1 } from '../constants.js';
+import { BASE_URL, EXPLORER_SERVICE_BASE_URL } from '../constants.js';
 const axios = require('axios');
 
 const errorHandler = (error) => {
@@ -27,9 +27,9 @@ const requestClient = () => {
   return instance;
 }
 
-const requestClientV1 = () => {
+const requestExplorerServiceV1 = () => {
   const defaultOptions = {
-    baseURL: BASE_URL_V1,
+    baseURL: EXPLORER_SERVICE_BASE_URL,
     headers: {
       'Content-Type': 'application/json',
     },
@@ -44,4 +44,4 @@ const requestClientV1 = () => {
 
 export default requestClient();
 
-export { requestClientV1 };
+export { requestExplorerServiceV1 };
--- a/src/components/TokenNFTPreview.js
+++ b/src/components/TokenNFTPreview.js
@@ -7,18 +7,18 @@ const TokenNFTPreview = (props) => {
     setToken(props.token);
   }, [props.token]);
 
-  if (!token.nft) {
+  if (!token.meta.nft) {
     return null;
   }
 
   let media;
 
-  if (token.nft.type === 'image') {
-    media = <img src={token.nft.file} width="100%" height="100%" />;
+  if (token.meta.nft.type === 'image') {
+    media = <img src={token.meta.nft.file} width="100%" height="100%" alt="NFT Preview" />;
   } else {
     media = (
       <video controls controlsList="nodownload noremoteplayback" disablePictureInPicture>
-        <source src={token.nft.file} type="video/mp4" />
+        <source src={token.meta.nft.file} type="video/mp4" />
         Your browser does not support html video tag.
       </video>
     )
--- a/src/components/TokenNFTPreview.js
+++ b/src/components/TokenNFTPreview.js
@@ -4,6 +4,7 @@ const TokenNFTPreview = (props) => {
   const [token, setToken] = useState(props.token);
 
   useEffect(() => {
+    // ensure data binding with props.token re rendering when changed
     setToken(props.token);
   }, [props.token]);
 
--- a/src/components/TokenNFTPreview.js
+++ b/src/components/TokenNFTPreview.js
@@ -1,0 +1,37 @@
+import React, { useState, useEffect } from 'react';
+
+const TokenNFTPreview = (props) => {
+  const [token, setToken] = useState(props.token);
+
+  useEffect(() => {
+    setToken(props.token);
+  }, [props.token]);
+
+  if (!token.nft) {
+    return null;
+  }
+
+  let media;
+
+  if (token.nft.type === 'image') {
+    media = <img src={token.nft.file} width="100%" height="100%" />;
+  } else {
+    media = (
+      <video controls controlsList="nodownload noremoteplayback" disablePictureInPicture>
+        <source src={token.nft.file} type="video/mp4" />
+        Your browser does not support html video tag.
+      </video>
+    )
+  }
+
+  return (
+    <div className="d-flex flex-column token-nft-preview">
+      <p><strong>NFT preview</strong></p>
+      <figure class="figure flex-fill p-4 d-flex align-items-center justify-content-center">
+        { media }
+      </figure>
+    </div>        
+  );
+}
+
+export default TokenNFTPreview;
--- a/src/components/TokenNFTPreview.js
+++ b/src/components/TokenNFTPreview.js
@@ -13,7 +13,7 @@ const TokenNFTPreview = (props) => {
 
   let media;
 
-  if (token.meta.nft.type === 'image') {
+  if (token.meta.nft.type && token.meta.nft.type.toUpperCase() === 'IMAGE') {
     media = <img src={token.meta.nft.file} width="100%" height="100%" alt="NFT Preview" />;
   } else {
     media = (
--- a/src/constants.js
+++ b/src/constants.js
@@ -6,6 +6,7 @@
  */
 
 export const BASE_URL = process.env.REACT_APP_BASE_URL || "https://node.explorer.hathor.network/v1a/";
+export const BASE_URL_V1 = process.env.REACT_APP_BASE_URL_V1 || "https://explorer-service.hathor.network/v1/";
 
 let tmp_ws_url = process.env.REACT_APP_WS_URL || "wss://node.explorer.hathor.network/v1a/ws/";
 if (!(tmp_ws_url.startsWith('ws:') || tmp_ws_url.startsWith('wss:'))) {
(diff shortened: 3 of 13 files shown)


src/screens/TokenDetail.js
```diff
@@ -199,17 +227,39 @@ class TokenDetail extends React.Component {
...
             </span> 
-            <QRCode size={200} value={configurationString} />
-            <a className="mt-2" onClick={(e) => this.downloadQrCode(e)} download={`${this.state.token.name} (${this.state.token.symbol}) - ${configurationString}`} href="true" ref="downloadLink">Download <i className="fa fa-download ml-1" title="Download QRCode"></i></a>
+            <a className="mt-2" onClick={(e) => this.downloadQrCode(e)} download={`${this.state.token.name} (${this.state.token.symbol}) - ${configurationString}`} href="true" ref="downloadLink">
+              Download
+              <i className="fa fa-download ml-1 float-right" title="Download QRCode"></i>
+            </a>
           </div>
+          <>
+            { this.state.nft ?
```
andreabadesso, line 247: Isn't there a better way to write this code block?

I find ternary operators on jsx confusing to understand

Maybe create another component to render the NFT

What do you think?

andreabadesso (changes requested):

src/screens/TokenDetail.js
giovanecosta: 100% agreed. Just did it.

andreabadesso (dismissed):

src/screens/TokenDetail.js
```diff
@@ -187,10 +215,36 @@ class TokenDetail extends React.Component {
...
+    const renderNftPreview = () => {
+      if (this.state.nft) {
+
+        let media;
+
+        if (this.state.nft.type === 'image') {
+          media = <img src={this.state.nft.file} width="100%" height="100%" />;
+        } else {
+          media = (
+            <video controls>
```
pedroferreira1, line 227: What do we have in the three dots in the video player? I think we should at least remove the download option. We could leave volume, full screen but (even though we know it can be downloaded anyway) we should try to hide this download button. This below should work:

```
controlsList="nodownload"
```

pedroferreira1 (dismissed):

src/screens/TokenDetail.js
msbrogli: Why should we remove the download option? I feel that people are free to download the asset behind an NFT.

src/screens/TokenDetail.js
```diff
@@ -74,13 +74,16 @@ class TokenDetail extends React.Component {
...
    */
   fetchNftList() {
-    fetch(helpers.nftTokenListUrl()).then((response) => {
-      response.json().then((data) => {
-        this.setState({
-          nft: data.find(obj => obj.id === this.state.paramUID)
+    const nftTokenListUrl = helpers.nftTokenListUrl();
+    if (nftTokenListUrl) {
+      fetch(helpers.nftTokenListUrl()).then((response) => {
+        response.json().then((data) => {
```
andreabadesso, line 80: Shouldn't we handle error here? Also, won't the app will crash if the request was successful but returned invalid data?

andreabadesso (changes requested):

src/screens/TokenDetail.js
giovanecosta: Sure! But as we don't have the real urls yet, i don't think we have to care about it now. And probably the only thing we can do on errors is log.

The side effect is just the `state` not being changed. So nothing happens, no media showing for the token if it have any.

andreabadesso: Ok but what I mean is that if `data` (returned from the API) is not an array, `data.find` will throw an error that is not being handled and the explorer screen will be white, right?

giovanecosta: Will not crash, just throw an exception on console. Anyway, it's protected from crash now: https://github.com/HathorNetwork/hathor-explorer/pull/54/commits/6a8ce3f72b531d5792e9966e7cc57b41b71fb103#diff-b6137568dc9d2ef5b53c1a08a45e9d9ee38a212a149533ce384c5858fc1103d8R81

src/screens/TokenDetail.js
pedroferreira1: I agree that people are free but (i) I think the player is small and have many buttons, thus removing the 3 dots icon would be better imo for UI and (ii) the idea of the NFT detail in the explorer is to show, not download (if the idea was to have a download option the design would have it).

msbrogli: I agree. We can move forward without a download button.

src/constants.js
```diff
@@ -26,21 +26,25 @@ export const DASHBOARD_TX_COUNT = 6;
...
-]
+];
 
 export const TESTNET_GENESIS_TX = [
   '00e161a6b0bee1781ea9300680913fb76fd0fac4acab527cd9626cc1514abdc9',
   '00975897028ceb037307327c953f5e7ad4d3f42402d71bd3d11ecb63ac39f01a',
-]
+];
+
+export const MAINNET_NFT_TOKEN_LIST_URL = '';
```
msbrogli, line 45: Shouldn't we remove this config?

src/constants.js
```diff
@@ -26,21 +26,25 @@ export const DASHBOARD_TX_COUNT = 6;
...
+
+export const TESTNET_NFT_TOKEN_LIST_URL = '';
```
msbrogli, line 47: Shouldn't we remove this config?

src/screens/TokenDetail.js
```diff
@@ -67,6 +70,24 @@ class TokenDetail extends React.Component {
     });
   }
 
+  /**
+   * Fetch NFT info for custom NFT Tokens
+   */
+  fetchNftList() {
```
msbrogli, line 76: I don't feel this method should be inside the screen. Let's move it somewhere else so we can easily test it. And I also think the best method is `getTokenMetadata(...)` which returns a promise that will be resolved for a json with all metadata available. In this case, we can have another method in this screen called `isNFT(...)` that receives this json and returns a boolean.

src/screens/TokenDetail.js
msbrogli: I'm afraid we won't use this method anymore. Take a look at my other comment, please.

src/screens/TokenDetail.js
```diff
@@ -187,10 +220,36 @@ class TokenDetail extends React.Component {
       )
     }
 
+    const renderNftPreview = () => {
+      if (this.state.nft) {
```
msbrogli, line 224: The whole method is inside this if statement. I feel we should change to:

```javascript
const renderNftPreview = () => {
  if (!this.state.nft) {
    return;
  }
  :
  (here goes the code)
  :
}
```

src/utils/helpers.js
```diff
@@ -192,6 +193,19 @@ const helpers = {
     return TESTNET_GENESIS_BLOCK.includes(hash);
   },
 
+  /**
+   * Get NFT toke list url
+   *
+   * @returns {string} url of token list
+   */
+  nftTokenListUrl() {
```
msbrogli, line 201: I feel we can remove it.

src/screens/TokenDetail.js
```diff
@@ -166,15 +187,27 @@ class TokenDetail extends React.Component {
     const renderTokenInfo = () => {
       return (
         <div className="token-general-info">
-          <p className="mb-2"><strong>UID: </strong>{this.state.token.uid}</p>
```
msbrogli, line 169: Isn't it better to have two components? One for regular tokens and a separate one for NFT tokens? I mean, the more we want to change the NFT screen, the more we have to guarantee we are not messing with the regular token screen.

giovanecosta: I made it! And i've also broke into three components (one for each section), so we have a NFT version and a non-NFT version that uses these components.

src/screens/TokenDetail.js
giovanecosta: Agreed. But i moved it to a new component.

src/screens/TokenDetail.js
```diff
@@ -21,85 +21,66 @@ import { TX_COUNT } from '../constants';
...
-      this.showSuccess('Configuration string copied to clipboard!');
-    }
+  updateTokenMetadata = (id) => {
+    tokenApi.getMetadata(id).then((data) => {
+      if (data) {
+        this.setState((oldState) => {
+          return {
+            token: {
+              ...oldState.token,
+              uid: id,
```
msbrogli, line 77: Why do you set the `uid` again?

src/screens/TokenDetail.js
```diff
@@ -21,85 +21,66 @@ import { TX_COUNT } from '../constants';
...
    * token {Object} selected token data
-   * successMessage {string} success message to show
    * errorMessage {string} error message to show
-   * totalSupply {number} Token total supply
-   * canMint {boolean} If this token can still be minted
-   * canMelt {boolean} If this token can still be melted
-   * paramUID {string} UID of token in the URL
    * transactions {Array} Array of transactions for the token
    */
   state = {
```
msbrogli, line 27: We should have at least all keys that will be eventually created here.

src/api/axiosInstance.js
```diff
@@ -19,10 +23,25 @@ const requestClient = () => {
...
   instance.interceptors.response.use((response) => {
     return response;
-  }, (error) => {
-    console.log("ERROR RESPONSE");
-  });
+  }, errorHandler);
+  return instance;
+}
+
+const requestClientV1 = () => {
```
msbrogli, line 30: Rename to `requestExplorerServiceV1(...)`?

msbrogli (dismissed):

src/screens/TokenDetail.js
```diff
@@ -21,85 +21,66 @@ import { TX_COUNT } from '../constants';
 class TokenDetail extends React.Component {
   /**
    * token {Object} selected token data
```
pedroferreira1, line 23: Can you  describe here the expected keys of the `Object`? When we write 'token data' is usually means an object with {uid, name, symbol} but you are adding lots of information on it right now, right?

src/screens/TokenDetail.js
pedroferreira1: Maybe it's because `updateTokenMetadata` and `updateTokenInfo` are async calls, then either of them could finish first. And if he calls `setState` without a `this.state.token.uid` it will break the component jsx, is that right?

I like having both API calls async to be honest, maybe we should set the token uid on state before the API calls.

pedroferreira1 (approved):

src/screens/TokenDetail.js
giovanecosta: There's a bug on `Transactions` component that it is not properly reactive. If the `token` is set without `uid` it makes a resquest to retrieve the list with `undefined` and never rerender again.

As the request to metadata service can finish first than the tokendetails, we have to set the `uid` when setting the token.

src/constants.js
```diff
@@ -6,6 +6,7 @@
  */
 
 export const BASE_URL = process.env.REACT_APP_BASE_URL || "https://node.explorer.hathor.network/v1a/";
+export const BASE_URL_V1 = process.env.REACT_APP_BASE_URL_V1 || "https://explorer-service.hathor.network/v1/";
```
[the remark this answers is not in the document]
giovanecosta: Yeah, in fact it is a better name.

src/screens/TokenDetail.js
giovanecosta: All those keys will be inside `token` object now

src/api/axiosInstance.js
giovanecosta: Sure, make sense after renaming the constant

src/screens/TokenDetail.js
giovanecosta: Yes, @msbrogli said it above and i agree too.

src/screens/TokenDetail.js
pedroferreira1: You can always substitute

```
setTokenId(id) {
  this.updateTokenInfo(id);
  this.updateTokenMetadata(id); 
}
```

with

```
setTokenId(id) {
  this.setState({ token: { uid: id } }, () => {
    this.updateTokenInfo(id);
    this.updateTokenMetadata(id); 
  });
}
```

Then you won't need to set the uid again in both responses. But I am also okay to keep like it's coded right now.

pedroferreira1 (approved):

src/components/TokenNFTPreview.js
```diff
@@ -0,0 +1,37 @@
+import React, { useState, useEffect } from 'react';
+
+const TokenNFTPreview = (props) => {
+  const [token, setToken] = useState(props.token);
+
+  useEffect(() => {
```
andreabadesso, line 6: Why do we need the `useEffect` hook? You could just access `props.token` directly or am I missing something?

andreabadesso (changes requested):

src/components/TokenNFTPreview.js
giovanecosta: When token is first passed, it doesn't have all attributes set. `useEffect` is needed to update the state when `token` change:

https://stackoverflow.com/a/59289634

andreabadesso: I see! 

Can you just add a comment to explain that you are using the useEffect to force a render?

andreabadesso (approved):

pedroferreira1 (approved):

msbrogli (approved):

andreabadesso (approved):

PR #60, opened 2021-05-31
ci: automate deployment

luislhl:
This PR automates the deployment of the service.

The deployment flow will be:
- Commits to master will trigger a deploy to the Testnet Explorer
- Creation of a new tag will trigger a deploy to the Mainnet Explorer

There is also a version checking during deploy, to make sure we don't forget to bump versions in the needed files, and that the git tag version also matches those files.

### Observations
- I also had to make changes in code to remove some warnings that were breaking the build. We could also just disable this behavior of breaking on warnings, but they seemed easily fixable, so I chose to fix them.

### TODO
- [x] Add secrets in repo


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,58 @@
...
+    if: github.ref == 'refs/heads/master'
+    needs: dependencies
+    runs-on: ubuntu-latest
+    steps:
+      - name: Build
+        run: |
+          npm run build
+        env:
+          REACT_APP_BASE_URL: https://node.explorer.hathor.network/v1a/
+          REACT_APP_WS_URL: wss://node.explorer.hathor.network/v1a/
```
luislhl, line 30: This needs to be fixed with the right value

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,90 @@
...
+        run: |
+          tar -xf node_modules.tar
+          npm run build
+        env:
+          REACT_APP_BASE_URL: https://node1.foxtrot.testnet.hathor.network/v1a/
+          REACT_APP_WS_URL: wss://node1.foxtrot.testnet.hathor.network/v1a/ws/
+      - name: Deploy Testnet Explorer
+        run: |
+          make check_version
+          aws s3 sync --delete ./build/ s3://hathor-testnet-foxtrot-public-explorer
```
pedroferreira1, line 60: Maybe we could create a make command for this script if I want to manually trigger an update on the testnet.  We can also create the make command for the mainnet.

Makefile
```diff
@@ -8,6 +8,10 @@ build:
 	REACT_APP_BASE_URL=$(REACT_APP_BASE_URL) \
 	npm run build
 
-.PHONY: s3_upload
```
pedroferreira1, line 11: Now I am seeing that we already have this make command. Why don't we keep this, change the name to `testnet_deploy`, create another one `mainnet_deploy` and use the make instead of having the script directly on github actions?

pedroferreira1 (dismissed):

.github/workflows/deploy.yml
luislhl: That's ok for me, will do it.

Makefile
luislhl: Done

pedroferreira1 (dismissed):

msbrogli (dismissed):

pedroferreira1 (approved):

pedroferreira1 (approved):

msbrogli (approved):

PR #71, opened 2021-06-09
feat: add analytics

giovanecosta:


src/constants.js
```diff
@@ -65,3 +65,5 @@ export const HATHOR_TOKEN_CONFIG = {'name': 'Hathor', 'symbol': 'HTR', 'uid': '0
  * Hathor token default index
  */
 export const HATHOR_TOKEN_INDEX = 0;
+
+export const GTM_ID = 'GTM-MJVX6BG';
```
luislhl, line 69: Will we be sending both testnet and mainnet events using the same GTM_ID?
Shouldn't they have one ID each?

giovanecosta: Changed to env var

src/constants.js
```diff
@@ -65,3 +65,5 @@ export const HATHOR_TOKEN_CONFIG = {'name': 'Hathor', 'symbol': 'HTR', 'uid': '0
  * Hathor token default index
  */
 export const HATHOR_TOKEN_INDEX = 0;
+
+export const GTM_ID = process.env.REACT_APP_GTM_ID;
```
pedroferreira1, line 69: I don't think we should have a `GTM_ID` for testnet explorer.

How `TagManager.initialize({ gtmId: GTM_ID })` will behave in case `GTM_ID` is an empty string or null?

Side note: change the target branch to `dev`

src/components/TokenNFTPreview.js
```diff
@@ -1,24 +1,38 @@
...
 
 const TokenNFTPreview = (props) => {
   const [token, setToken] = useState(props.token);
 
   useEffect(() => {
     // ensure data binding with props.token re rendering when changed
     setToken(props.token);
+    TagManager.dataLayer({
+      dataLayer: {
+        'event': 'LoadNFTPreview'
```
pedroferreira1, line 12: Can we add here the token ID in the event? So we can know which token is being loaded more? Or this event already group by token because of the URL?

src/components/TokenNFTPreview.js
```diff
@@ -1,24 +1,38 @@
...
   }, [props.token]);
 
   if (!token.meta.nft) {
     return null;
   }
 
+  const onPLayVideo = () => {
+    TagManager.dataLayer({
+      dataLayer: {
+        'event': 'PlayNFTVideo'
```
pedroferreira1, line 24: Same comment as above about the token ID

src/constants.js
pedroferreira1: Maybe we could have only mainnet tag

src/components/TokenNFTPreview.js
giovanecosta: Yes we can. Just checked if the page url is sent with the event, and it is. However for better use i've decided to add the `token.uid` as well.

pedroferreira1 (approved):

src/constants.js
luislhl, line 69: I think it's becoming difficult to know exactly which env vars this project expects. Maybe we should at least keep them all together in this file (close to each other).

But the main thing here is that now this env var is required, right?

But for development, one shouldn't have to define this to make the project run in his/her machine. We should have some mechanism in place to make this env var optional for development.

luislhl (approved):

PR #72, opened 2021-06-09
fix: add margin to boxes on mobile

giovanecosta:
# Before
image

# After
image


pedroferreira1 (dismissed):
Just two things:

1. PR to dev always
2. Would be great to have a screenshot

pedroferreira1 (approved):

Issue #73, opened 2021-06-16
Select Node on Network Page


giovanecosta:
Change Network Page to use `/node` endpoint of `explorer-service` and select different nodes.

PR #75, opened 2021-06-16
feat(p2p): new network page

giovanecosta:
https://github.com/HathorNetwork/hathor-explorer/issues/73

image
image

https://user-images.githubusercontent.com/698586/123879286-0ff32400-d917-11eb-8812-f105d9243d42.mp4

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -111,7 +111,7 @@ class Root extends React.Component {
               <NavigationRoute exact path="/transactions" component={TransactionList} />
               <NavigationRoute exact path="/blocks" component={BlockList} />
               <NavigationRoute exact path="/dag" component={Dag} />
-              <NavigationRoute exact path="/network/:peer_id?" component={PeerAdmin} />
+              <NavigationRoute exact path="/network/:peerId?" component={PeerAdmin} />
               <NavigationRoute exact path="/statistics" component={Dashboard} />
               <NavigationRoute exact path="/token_detail/:tokenUID" component={TokenDetail} />
               <NavigationRoute exact path="/tokens" component={TokensList} />
--- a/src/App.js
+++ b/src/App.js
@@ -111,7 +111,7 @@ class Root extends React.Component {
               <NavigationRoute exact path="/transactions" component={TransactionList} />
               <NavigationRoute exact path="/blocks" component={BlockList} />
               <NavigationRoute exact path="/dag" component={Dag} />
-              <NavigationRoute exact path="/network" component={PeerAdmin} />
+              <NavigationRoute exact path="/network/:peer_id?" component={PeerAdmin} />
               <NavigationRoute exact path="/statistics" component={Dashboard} />
               <NavigationRoute exact path="/token_detail/:tokenUID" component={TokenDetail} />
               <NavigationRoute exact path="/tokens" component={TokensList} />
--- a/src/api/networkApi.js
+++ b/src/api/networkApi.js
@@ -5,26 +5,18 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import requestClient from './axiosInstance';
+import { requestExplorerServiceV1 } from './axiosInstance';
 
 const networkApi = {
   getPeers() {
-    requestClient.interceptors.request.use(config => { 
-      config.baseURL = 'http://localhost:3001/dev/';
-      return config; 
-    });
-    return requestClient.get(`node`).then((res) => {
+    return requestExplorerServiceV1().get(`node`).then((res) => {
       return res.data
     }, (res) => {
       throw new Error(res.data.message);
     });
   },
   getPeer(hash) {
-    requestClient.interceptors.request.use(config => { 
-      config.baseURL = 'http://localhost:3001/dev/';
-      return config; 
-    });
-    return requestClient.get(`node/${hash}`).then((res) => {
+    return requestExplorerServiceV1().get(`node/${hash}`).then((res) => {
       return res.data
     }, (res) => {
       throw new Error(res.data.message);
--- a/src/api/networkApi.js
+++ b/src/api/networkApi.js
@@ -9,7 +9,22 @@ import requestClient from './axiosInstance';
 
 const networkApi = {
   getPeers() {
-    return requestClient.get(`status`).then((res) => {
+    requestClient.interceptors.request.use(config => { 
+      config.baseURL = 'http://localhost:3001/dev/';
+      return config; 
+    });
+    return requestClient.get(`node`).then((res) => {
+      return res.data
+    }, (res) => {
+      throw new Error(res.data.message);
+    });
+  },
+  getPeer(hash) {
+    requestClient.interceptors.request.use(config => { 
+      config.baseURL = 'http://localhost:3001/dev/';
+      return config; 
+    });
+    return requestClient.get(`node/${hash}`).then((res) => {
       return res.data
     }, (res) => {
       throw new Error(res.data.message);
--- a/src/api/networkApi.js
+++ b/src/api/networkApi.js
@@ -8,17 +8,17 @@
 import { requestExplorerServiceV1 } from './axiosInstance';
 
 const networkApi = {
-  getPeers() {
+  getPeerList() {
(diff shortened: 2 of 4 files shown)


pedroferreira1 (changes requested):

src/api/networkApi.js
```diff
@@ -5,16 +5,23 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import requestClient from './axiosInstance';
+import { requestExplorerServiceV1 } from './axiosInstance';
 
 const networkApi = {
   getPeers() {
```
msbrogli, line 11: I'd rename it to `getPeerList()` because `getPeers` might easily be miswritten as `getPeer` and vice versa.

src/components/Network.js
```diff
@@ -33,27 +34,45 @@ class Network extends React.Component {
...
 
     this.loadData = this.loadData.bind(this);
   }
 
   componentDidMount() {
-    this.loadData();
-    this.loadTimer = setInterval(this.loadData, 1000);
+    const { match: { params } } = this.props;
+
+    networkApi.getPeers().then((peers) => {
```
msbrogli, line 48: What if this request fails? What happens? Shouldn't we show an error message and keep trying every N seconds?

src/components/Network.js
```diff
@@ -33,27 +34,45 @@ class Network extends React.Component {
...
+      this.setState({peers}, () => {
+        let peer_id = params.peer_id;
+        if (!peer_id) {
+          peer_id = peers[0];
+        }
+        this.props.history.push(`/network/${this.state.peers[0]}`);
```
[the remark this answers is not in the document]
msbrogli: I agree with the refactor.

src/components/Network.js
msbrogli, line 48: Refactor and move this piece of code to a separate method that will be called from here.

src/components/Network.js
```diff
@@ -33,27 +34,45 @@ class Network extends React.Component {
...
+        this.setState({peer_id}, () => {
```
msbrogli, line 55: Add comment explaining that you will load the data immediately and also schedule an update every second.

src/components/Network.js
```diff
@@ -33,27 +34,45 @@ class Network extends React.Component {
...
+    this.props.history.push(`/network/${peer_id}`);
+    this.setState({peer_id}, () => { this.loadData() });
   }
 
   loadData() {
     if (this.state.isLoading) return;
 
     this.setState({ isLoading: true }, () => {
-      networkApi.getPeers().then((peers) => {
+      networkApi.getPeer(this.state.peer_id).then((peer) => {
```
msbrogli, line 73: What if this request fails?

src/components/Network.js
```diff
@@ -149,14 +173,14 @@ class Network extends React.Component {
               Uptime: {dateFormatter.uptimeFormat(conn.uptime)}<br />
               Version: {conn.app_version}<br />
               Address: {conn.address}<br />
-              Entrypoints: {peer.entrypoints.join(", ")}<br />
+              Entrypoints: { /* peer.entrypoints.join(", ") */ }<br />
```
msbrogli, line 176: Why did you comment it out? I think we should show the peer's entrypoints.

src/components/Network.js
```diff
@@ -184,11 +208,31 @@ class Network extends React.Component {
...
 
+    const renderControls = () => {
+      if (!this.state.loaded) {
+        return null;
+      }
+
+      return (
+        <div className="form-inline mb-3">
+          <label>
+            Peer PoV:
```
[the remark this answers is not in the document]
msbrogli: I agree. And I feel we should add a "?" icon with a brief explanation. Suggestion: "Select a peer to check its network status.".

src/App.js
```diff
@@ -112,7 +112,7 @@ class Root extends React.Component {
           <NavigationRoute exact path="/transactions" component={TransactionList} />
           <NavigationRoute exact path="/blocks" component={BlockList} />
           <NavigationRoute exact path="/dag" component={Dag} />
-          <NavigationRoute exact path="/network" component={PeerAdmin} />
```
msbrogli, line 115: Will the url `/network` (without the last slash) work?

msbrogli (changes requested):

src/components/Network.js
```diff
@@ -81,25 +100,26 @@ class Network extends React.Component {
   }
 
   buildSyncDataPolyfill(conn) {
```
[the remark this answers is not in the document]
giovanecosta: Yes, it will.

src/components/Network.js
```diff
@@ -33,27 +33,49 @@ class Network extends React.Component {
...
+  }
+
+  loadPeers() {
+    const { match: { params } } = this.props;
+
+    networkApi.getPeerList().then((peers) => {
+      this.setState({peers}, () => {
+        let peerId = peers.find(target => target === params.peerId);
+        if (!peerId) {
+          peerId = peers[0];
```
pedroferreira1, line 54: Should we handle the case when the API returns an empty array of peers? It should throw an error here, right? I know it shouldn't happen but I am thinking if we should do something to protect this. What do you think?

src/components/Network.js
pedroferreira1: Add in this issue to add back the entrypoints on explorer when it's added there

pedroferreira1 (dismissed):

pedroferreira1:
Can you add a video with the change peer flow?

src/components/Network.js
giovanecosta: Done

src/components/Network.js
giovanecosta: I'm not sure where is my response for @msbrogli about what happens when the request fail. 

What happen is: the Loading will be shown until page reload. As we don't have a good way to show errors nowadays, i think we can wait for refactors that will be made soon.

src/components/Network.js
giovanecosta: We don't have a good way to handle, or show errors in general for Explorer yet. With the way we use Axios nowadays, it's even more tricky. I'll open an issue for that.

For this case It will retry every second and show the Loading component while it's not succeeded

src/components/Network.js
msbrogli: I don't get it. Do we filter out the full-node's response? Anyway, I think we should make this code work when the information is correctly returned instead of this commenting this out. Maybe the API Client can fix it putting an empty list there is nothing is return.

src/components/Network.js
msbrogli: Ok. Let's create another issue to add an error screen, please.

msbrogli (changes requested):
You have removed a code because of a missing information in the wallet service API. I just put my thoughts in the discussion. It is the last point to be fixed and I'll approve it.

src/components/Network.js
giovanecosta: I agree, changed to work with or without entrypoints coming

src/components/Network.js
pedroferreira1: I was not talking about when a request fails. I was saying that the request may succeed and returns an empty array. You are assuming the array will always have at least one element and doing `peers[0]`. As we don't have a good way to show errors, you can at least write a TODO (Check if array is empty and show error)

pedroferreira1 (approved):

src/components/Network.js
msbrogli: Great. Thanks!

msbrogli (approved):

src/components/Network.js
giovanecosta: It's not exactly the same thing but is the same behavior: with an empty list `peers[0]` will result in `undefined`, so, the next request will pass `undefined` as id and the request will fail. The best we can do here, for now, is try to fetch node ids again instead of just go ahead. But the behavior will remain the same, except for the very edge case when the peer list is empty when user load the page and start to have items while the page is open.

src/components/Network.js
giovanecosta: @msbrogli created #85

PR #77, opened 2021-06-17
feat(tx): add NFT audio preview

giovanecosta:
image

Mobile:
image

### DIFF
--- a/src/components/TokenNFTPreview.js
+++ b/src/components/TokenNFTPreview.js
@@ -17,13 +17,13 @@ const TokenNFTPreview = (props) => {
         }
       });
     }
-  }, [props.token]);
+  }, [props.token, token.uid]);
 
   if (!token.meta.nft) {
     return null;
   }
 
-  const onPLayMedia = () => {
+  const onPlayMedia = () => {
     TagManager.dataLayer({
       dataLayer: {
         event: 'PlayNFTMedia',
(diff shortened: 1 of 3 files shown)


src/components/TokenNFTPreview.js
```diff
@@ -21,26 +23,49 @@ const TokenNFTPreview = (props) => {
     return null;
   }
 
-  const onPLayVideo = () => {
+  const onPLayMedia = () => {
```
pedroferreira1, line 26: This method name should be `onPlayMedia`, right? I think the L was not supposed to be uppercase

pedroferreira1 (dismissed):
Missing screenshot

src/components/TokenNFTPreview.js
msbrogli: Did it work during tests?

src/utils/helpers.js
```diff
@@ -201,6 +201,20 @@ const helpers = {
...
+   *
+   * @param {string} file File name
+   *
+   * @return {string} extension of the file
+   * @memberof Helpers
+   * @inner
+   */
+  getFileExtension(file) {
+    const parts = file.split('.');
+    return parts[parts.length - 1];
```
msbrogli, line 217: If the file has no extension at all (e.g. `myfile`), it will return the base name instead of the extension. Maybe add the following?
```nodejs
if (parts.length == 1) {
  return '';
}
```

msbrogli (dismissed):

src/components/TokenNFTPreview.js
pedroferreira1: Probably yes, the execution call is also using L. It's just to keep the pattern.

src/utils/helpers.js
giovanecosta: Sure! I think we can still improve given that the file is an URI. But don't think we have to care about now.

src/components/TokenNFTPreview.js
giovanecosta: Ok, my fingers betrayed me in conspiracy with code completion, haha. Fixed.

src/utils/helpers.js
msbrogli: If things get more complicated, maybe we should use a library that parses URI. But it doesn't seem necessary now.

src/components/TokenNFTPreview.js
msbrogli: image

msbrogli (approved):

pedroferreira1 (approved):

PR #80, opened 2021-06-23
feat: add cookie consent message

giovanecosta:
# Desktop
image

# Mobile
image

Added some explanation here: 
https://github.com/HathorNetwork/internal-issues/issues/29

### DIFF
--- a/src/components/GDPRConsent.js
+++ b/src/components/GDPRConsent.js
@@ -1,0 +1,46 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React, { useEffect } from 'react';
+
+import { GTM_ID } from '../constants';
+import CookieConsent, { getCookieConsentValue }  from "react-cookie-consent";
+import TagManager from 'react-gtm-module';
+
+
+const GDPRConsent = () => {
+  
+  const onAccept = () => {
+    // Nothing yet
+  };
+  
+  useEffect(() => {
+    // Just to ensure that when we initialize GTM, this message will be shown
+    TagManager.initialize({ gtmId: GTM_ID });
+
+    const isConsent = getCookieConsentValue();
+    if (isConsent === "true") {
+      onAccept();
+    }
+  }, []);
+
+  return (
+    <div className="cookie-consent-wrapper">
+      <CookieConsent
+        onAccept={onAccept}
+        disableButtonStyles={true}
+        buttonWrapperClasses="mx-auto"
+        buttonClasses="btn btn-hathor m-3"
+        >
+        We use cookies to improve your user experience. By continuing onto our website, you agree.
+      </CookieConsent>
+    </div>
+  )
+}
+
+export default GDPRConsent;
+ 
--- a/src/components/GDPRConsent.js
+++ b/src/components/GDPRConsent.js
@@ -19,7 +19,7 @@ const GDPRConsent = () => {
   };
   
   useEffect(() => {
-    // Just to ensure that when we initialize GTM, this message will be shown
+    // Just to ensure that when we initialize GTM, the message on this component will be shown
     TagManager.initialize({ gtmId: GTM_ID });
 
     const isConsent = getCookieConsentValue();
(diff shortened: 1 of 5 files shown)


src/components/GDPRConsent.js
```diff
@@ -0,0 +1,46 @@
...
+  const onAccept = () => {
+    // Nothing yet
+  };
+  
+  useEffect(() => {
+    // Just to ensure that when we initialize GTM, this message will be shown
+    TagManager.initialize({ gtmId: GTM_ID });
+
+    const isConsent = getCookieConsentValue();
+    if (isConsent === "true") {
```
pedroferreira1, line 26: Does it return a string?

pedroferreira1 (dismissed):

src/components/GDPRConsent.js
```diff
@@ -0,0 +1,46 @@
...
+      onAccept();
+    }
+  }, []);
+
+  return (
+    <div className="cookie-consent-wrapper">
```
msbrogli, line 32: Why do we need this div?

src/components/GDPRConsent.js
```diff
@@ -0,0 +1,46 @@
...
+    // Nothing yet
+  };
+  
+  useEffect(() => {
+    // Just to ensure that when we initialize GTM, this message will be shown
+    TagManager.initialize({ gtmId: GTM_ID });
+
+    const isConsent = getCookieConsentValue();
+    if (isConsent === "true") {
+      onAccept();
```
msbrogli, line 27: What is it exactly doing since the `onAccept()` method does nothing?

src/components/GDPRConsent.js
```diff
@@ -0,0 +1,46 @@
...
+
+
+const GDPRConsent = () => {
+  
+  const onAccept = () => {
+    // Nothing yet
+  };
+  
+  useEffect(() => {
+    // Just to ensure that when we initialize GTM, this message will be shown
```
msbrogli, line 22: Which message?

src/components/GDPRConsent.js
```diff
@@ -0,0 +1,46 @@
...
+
+  return (
+    <div className="cookie-consent-wrapper">
+      <CookieConsent
+        onAccept={onAccept}
+        disableButtonStyles={true}
+        buttonWrapperClasses="mx-auto"
+        buttonClasses="btn btn-hathor m-3"
+        >
+        We use cookies to improve your user experience. By continuing onto our website, you agree.
```
msbrogli, line 39: Maybe we should do exactly as we do in our website: "This website uses cookies to ensure you get the best experience on our website. Learn more". The learn more is a link to https://hathor.network/terms-and-conditions/.

msbrogli (changes requested):

src/components/GDPRConsent.js
msbrogli: @giovanecosta @luislhl @pedroferreira1 What do you think?

src/components/GDPRConsent.js
giovanecosta: Yes! It comes from a cookie

src/components/GDPRConsent.js
giovanecosta: We don't. At first i was thinking about put an overlay to prevent users interaction until hits he consent button. But i changed this behavior. I'll remove.

src/components/GDPRConsent.js
giovanecosta: In case of we decide to put something that will run only if the user accept. A "consent cookie" is saved and next time the user access the site, we don't need to ask again, the consent is stored. So, we have to ensure the same behavior as he click on accept,

src/components/GDPRConsent.js
giovanecosta: The message on this component. I improved the comment.

src/components/GDPRConsent.js
giovanecosta: I've included the link but i still think that saying "By continue here you agree" thing is better.

pedroferreira1 (approved):

src/components/GDPRConsent.js
msbrogli: I thought the component would already do it. If it does, I think we must do it exactly as you described.

giovanecosta: I explain it here: https://github.com/HathorNetwork/internal-issues/issues/29

src/components/GDPRConsent.js
msbrogli: Ok. Thanks!

src/components/GDPRConsent.js
msbrogli: Ok. Just a minor fix: "... You agree with it.".

src/components/GDPRConsent.js
msbrogli: The way it is developed, will this message appear every refresh?

src/components/GDPRConsent.js
```diff
@@ -0,0 +1,36 @@
...
+import { GTM_ID } from '../constants';
+import CookieConsent from "react-cookie-consent";
+import TagManager from 'react-gtm-module';
+
+
+const GDPRConsent = () => {
+
+  useEffect(() => {
+    // Just to ensure that when we initialize GTM, the message on this component will be shown
+    TagManager.initialize({ gtmId: GTM_ID });
```
msbrogli, line 19: That's not the best place to do it, right? I think we can move forward but let's open an issue to get it fixed ASAP.

msbrogli (approved):

pedroferreira1 (approved):

src/components/GDPRConsent.js
giovanecosta: Yes! I'm already talinkg about it here: https://github.com/HathorNetwork/internal-issues/issues/29#issuecomment-867080613

PR #86, opened 2021-07-08
feat: show token metadata

giovanecosta:
# This is the first screen proposal.

When everybody approve this proposal and HathorNetwork/hathor-explorer-service#40 , i can add the request and finish.

## Scam Token
image

## Verified Token
image

## Others
image


src/components/token/TokenAlerts.js
```diff
@@ -0,0 +1,56 @@
...
+  }, [props.token]);
+
+  const verifiedAlert = () => {
+    if (!token.verified) {
+      return null;
+    }
+
+    return (
+      <div className="alert alert-info backup-alert" role="alert">
+        <i className="fa fa-check-circle mr-2" title="Token Verified"></i> This token is verified by Hathor team.
```
msbrogli, line 18: Maybe put the "Verified" icon next to the name of the token (just like in a Twitter account) and say "This a verified token. <a>Learn more</a>". We can use the same verified icon in the list of tokens inside a transaction.

src/components/token/TokenAlerts.js
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  const bannedAlert = () => {
+    if (!token.banned) {
+      return null;
+    }
+
+    return (
+      <div className="alert alert-danger backup-alert" role="alert">
+        <i className="fa fa-exclamation-triangle mr-2" title="Token Verified"></i> This token was marked as scam by Hathor team.
```
msbrogli, line 30: Maybe: "This token was banned and should not be trusted. <a>Learn more</a>". We should put Yan in the loop to check with our legal team what's the best language for this case.

src/components/token/TokenAlerts.js
```diff
@@ -0,0 +1,56 @@
+import React, { useState, useEffect } from 'react';
+
+
+const TokenAlerts = (props) => {
+  const [token, setToken] = useState(props.token);
+
+  useEffect(() => {
+    setToken({banned: true});
```
msbrogli, line 8: Testing code? I feel it will get the flags from the `props`, right?

src/components/token/TokenAlerts.js
```diff
@@ -0,0 +1,56 @@
...
+  }, [props.token]);
+
+  const verifiedAlert = () => {
+    if (!token.verified) {
```
msbrogli, line 12: `this.props.token.verified`?

msbrogli (changes requested):

src/components/token/TokenAlerts.js
giovanecosta: I think it will not be reactive that way.

src/components/token/TokenAlerts.js
giovanecosta: Ok, i'll get in touch with him

src/components/token/TokenAlerts.js
giovanecosta: I like it! I'll change and update

pedroferreira1: I agree with that.

I think even verified tokens must continue having the warning rectangle with the token UID message. The only difference should be the verified symbol besides the token symbol with a hover message.

obiyankenobi:
Did you discuss what are the requirements of a verified token? What's the process to do it? I understand this is more of a business decision.

@pedroferreira1 @msbrogli

giovanecosta:
@obiyankenobi updated and added some verified options

src/components/token/TokenAlerts.js
obiyankenobi: Will check it. Anyway, I feel that saying the token was banned sounds weird, as it can still be exchanged freely. We might want to tweak the message. But let's wait for what the legal team before we do that.

obiyankenobi:
> @obiyankenobi updated and added some verified options

I like option 2.

msbrogli:
I like option 2 as well. Besides the content of the message, is there anything else left to be done here? @giovanecosta

src/components/token/TokenAlerts.js
```diff
@@ -0,0 +1,34 @@
...
+
+  const bannedAlert = () => {
+    if (!token.banned) {
+      return null;
+    }
+
+    return (
+      <div className="alert alert-danger backup-alert" role="alert">
+        <i className="fa fa-exclamation-triangle mr-2" title="Token Verified"></i>
+        This token was banned and should not be trusted. <a href="http://google.com" >Learn more.</a>
```
msbrogli, line 19: Should we move this message to the Metadata? The only con I can see is that it won't be easy to have it translated in the future.

src/components/token/TokenMarkers.js
```diff
@@ -0,0 +1,49 @@
...
+  const bannedMarker = () => {
+    if (!token.banned) {
+      return null;
+    }
+
+    return (
+      <button className="info-hover-wrapper btn btn-link pl-2">
+        <i className="fa fa-exclamation-triangle fa-lg text-danger" title="Banned"></i>
+        <span className="subtitle info-hover-popover">
+          This token was banned and should not be trusted. <a href="http://google.com" >Learn more.</a>
```
msbrogli, line 35: Same thing here. And we should use a link to Hathor as example (not google).

src/components/token/TokenAlerts.js
msbrogli, line 19: Change the link to Hathor (instead of google).

msbrogli (changes requested):

src/components/token/TokenAlerts.js
pedroferreira1: And we need to repeat this message in all banned tokens metadata, right? If we decide to change a word, then we would need to change all metadatas, or am I missing something?

giovanecosta: We can have a key related to a i18n message instead of a message. However i think it's even better to have a default message and we can:
1. Override this message if the metadata have some different message
2. Have an extra message to have some additional information about the ban

For translating, the message field can be in the following format:
```
message: {
  "default": "Default message, probably in english",
  "pt_BR": "Mensagem padrão, provavelmente em inglês"
}
```

src/components/token/TokenAlerts.js
giovanecosta: Done

src/components/token/TokenAlerts.js
giovanecosta: I've created an issue here to discuss the message: https://github.com/HathorNetwork/hathor-explorer/issues/92

src/components/token/TokenAlerts.js
giovanecosta: changed

src/components/token/TokenMarkers.js
giovanecosta: changed

src/components/token/TokenAlerts.js
giovanecosta: changed

msbrogli: I understand that we might need to repeat the message in all banned tokens. On the other hand, we can have special cases where we want to place a different message.

But I agree that we should let the wallets and the public explorer decide which message to show. So, we need to put a code in the metadata (e.g., "REGULAR_BAN"). In the future, we can create other codes as needed.

msbrogli: In regards to translation, we should use PO files so we can use external translators. We have already built all tools to easily do that in `hathor-wallet` and `hathor-wallet-mobile`. We just have to port them to this repository.

src/components/token/TokenAlerts.js
```diff
@@ -0,0 +1,34 @@
...
+
+  const bannedAlert = () => {
+    if (!token.banned) {
+      return null;
+    }
+
+    return (
+      <div className="alert alert-danger backup-alert" role="alert">
+        <i className="fa fa-exclamation-triangle mr-2" title="Token Verified"></i>
+        This token was banned and should not be trusted. <a href="http://hathor.network" >Learn more.</a>
```
pedroferreira1, line 19: This is the message that is being discussed? Maybe we should leave a TODO only instead of this message. Besides that, would be better if the href link were for https instead of http, right?

src/components/token/TokenMarkers.js
```diff
@@ -0,0 +1,49 @@
...
+  const bannedMarker = () => {
+    if (!token.banned) {
+      return null;
+    }
+
+    return (
+      <button className="info-hover-wrapper btn btn-link pl-2">
+        <i className="fa fa-exclamation-triangle fa-lg text-danger" title="Banned"></i>
+        <span className="subtitle info-hover-popover">
+          This token was banned and should not be trusted. <a href="http://hathor.network" >Learn more.</a>
```
pedroferreira1, line 35: Same comment about the message and https in the URL

src/components/token/TokenMarkers.js
```diff
@@ -0,0 +1,49 @@
...
+  const verifiedMarker = () => {
+    if (!token.verified) {
+      return null;
+    }
+
+    return (
+      <button className="info-hover-wrapper btn btn-link pl-2">
+        <i className="fa fa-check-circle fa-lg text-info" title="Verified"></i>
+        <span className="subtitle info-hover-popover">
+          This is a verified token. <a href="http://hathor.network" >Learn more.</a>
```
pedroferreira1, line 20: Same comment about the https

src/components/token/TokenAlerts.js
giovanecosta: Moved to a const and added a todo

src/components/token/TokenMarkers.js
giovanecosta: Fixed

src/components/token/TokenMarkers.js
giovanecosta: this link (for banned) was removed

pedroferreira1 (dismissed):

src/components/Network.js
```diff
@@ -164,7 +164,7 @@ class Network extends React.Component {
         const { synced_percent, general_percent } = this.getSyncProgressPercent(sync_data);
         const sync_state_description = SyncStatesDescription[sync_data.state] || SyncStatesDescription[SyncStates.UNKNOWN];
 
-        const entrypoints = peer.entrypoints || [];
+        const entrypoints = conn.entrypoints || [];
```
msbrogli, line 167: Is it a bug fix?

src/components/token/TokenAlerts.js
```diff
@@ -0,0 +1,36 @@
...
+  useEffect(() => {
+    setToken(props.token);
+  }, [props.token]);
+
+  const bannedAlert = () => {
+    if (!token.banned) {
+      return null;
+    }
+
+    const bannedMessage = ''; // TODO: this message is being disscussed here: https://github.com/HathorNetwork/hathor-explorer/issues/92
```
msbrogli, line 16: This message is duplicated in `src/components/token/TokenMarkers.js`. We should move it somewhere else to keep it in only one place.

msbrogli (changes requested):

src/components/Network.js
giovanecosta: it's from another PR that it's here by accident. Removed.

src/components/token/TokenAlerts.js
giovanecosta: It will turn into something like `t('REGULAR_BAN')`.

pedroferreira1: The message is still duplicated, isn't it? You have this variable in both components.

pedroferreira1 (changes requested):

src/components/token/TokenAlerts.js
giovanecosta: Yes, they are the same translated message, the key is the same, they will come from the same i18n entry. If the message change, we won't have to change in both files. 

Of course, i can just remove this variable and put `t('REGULAR_BAN')` directly on the html but i've decided to let it explicit that it will changed. Also, i don't think it would have any advantages to move this to a different file now just to be alone and empty there.

giovanecosta: done!

msbrogli (approved):

pedroferreira1 (approved):

PR #87, opened 2021-07-12
Feat/show tx metadata

giovanecosta:
image

## Hover:

image

### DIFF
--- a/src/components/TxData.js
+++ b/src/components/TxData.js
@@ -18,6 +18,8 @@ import dateFormatter from '../utils/date';
 import txApi from '../api/txApi';
 import { BASE_URL, HATHOR_TOKEN_INDEX, HATHOR_TOKEN_CONFIG } from '../constants';
 import hathorLib from '@hathor/wallet-lib';
+import TxAlerts from './tx/TxAlerts';
+import TxMarkers from './tx/TxMarkers';
 
 
 /**
@@ -436,12 +438,6 @@ class TxData extends React.Component {
       );
     }
 
-    const renderGenesisMark = (hash) => {
-      if (helpers.isGenesisBlock(hash) || helpers.isGenesisTx(hash)) {
-        return <span className="text-info">[GENESIS]</span>
-      }
-    }
-
     const loadTxData = () => {
       return (
         <div className="tx-data-wrapper">
@@ -449,7 +445,7 @@ class TxData extends React.Component {
           <div><label>{hathorLib.helpers.isBlock(this.props.transaction) ? 'Block' : 'Transaction'} ID:</label> {this.props.transaction.hash}</div>
           <div className="d-flex flex-column flex-lg-row align-items-start mt-3 mb-3">
             <div className="d-flex flex-column align-items-start common-div bordered-wrapper mr-lg-3 w-100">
-              <div><label>Type:</label> {hathorLib.helpers.getTxType(this.props.transaction)} {renderGenesisMark(this.props.transaction.hash)}</div>
+              <div><label>Type:</label> {hathorLib.helpers.getTxType(this.props.transaction)} <TxMarkers tx={this.props.transaction} /></div>
               <div><label>Time:</label> {dateFormatter.parseTimestamp(this.props.transaction.timestamp)}</div>
               <div><label>Nonce:</label> {this.props.transaction.nonce}</div>
               <div><label>Weight:</label> {helpers.roundFloat(this.props.transaction.weight)}</div>
@@ -512,6 +508,7 @@ class TxData extends React.Component {
 
     return (
       <div>
+        <TxAlerts tx={this.props.transaction} />
         {loadTxData()}
         <HathorAlert ref="alertCopied" text="Copied to clipboard!" type="success" />
       </div>
--- a/src/components/tx/TxAlerts.js
+++ b/src/components/tx/TxAlerts.js
@@ -1,0 +1,30 @@
+import React, { useState, useEffect } from 'react';
+
+
+const TxAlerts = (props) => {
+  const [tx, setTx] = useState(props.tx);
+
+  useEffect(() => {
+    setTx({context: 'This is the 1,000,000th transaction.'});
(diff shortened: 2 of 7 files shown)


src/components/tx/TxAlerts.js
```diff
@@ -0,0 +1,30 @@
...
+    setTx({context: 'This is the 1,000,000th transaction.'});
+  }, [props.tx]);
+
+  const defaultAlert = () => {
+    if (!tx.context) {
+      return null;
+    }
+
+    return (
+      <div className="alert alert-info backup-alert" role="alert">
```
pedroferreira1, line 17: This `alert-info` or `alert-danger` could be defined depending on the metadata type as well.

src/components/TxData.js
```diff
@@ -512,6 +508,7 @@ class TxData extends React.Component {
 
     return (
       <div>
+        <TxAlerts tx={this.props.transaction} />
```
pedroferreira1, line 511: Commenting here to create a discussion thread.

I like the code in general but I think I don't like having the alert as an info. I believe when the tx has only info metadata, we should show only the marker, which I like a lot and leave only the success alert with 'This block/tx is valid'.

For any banned/scam txs I think we can add the red alert for more attention.

giovanecosta: > I believe when the tx has only info metadata, we should show only the marker

Like an `i` icon with the "this is the 1,000,000th" text on hover?

pedroferreira1: No, I mean only the TxMarker you created
image

msbrogli: I agree that the alerts must be different for each type. I think we should only use a whole line in the screen for specific alerts such as banned/scamming transactions.

src/components/tx/TxAlerts.js
```diff
@@ -0,0 +1,30 @@
+import React, { useState, useEffect } from 'react';
+
+
+const TxAlerts = (props) => {
+  const [tx, setTx] = useState(props.tx);
+
+  useEffect(() => {
+    setTx({context: 'This is the 1,000,000th transaction.'});
```
msbrogli, line 8: Debug code?

src/components/tx/TxAlerts.js
```diff
@@ -0,0 +1,30 @@
+import React, { useState, useEffect } from 'react';
+
+
+const TxAlerts = (props) => {
```
msbrogli, line 4: Why don't we use classes for this component?

src/components/tx/TxMarkers.js
```diff
@@ -0,0 +1,34 @@
+import React, { useState, useEffect } from 'react';
+import helpers from '../../utils/helpers';
+
+
+const TxMarkers = (props) => {
```
msbrogli, line 5: Why don't we use classes for this component?

src/components/tx/TxMarkers.js
```diff
@@ -0,0 +1,34 @@
...
+  }, [props.tx]);
+
+  const genesisMarker = () => {
+    if (!tx.genesis) {
+      return null;
+    }
+
+    return (
+      <button className="info-hover-wrapper btn btn-link pl-0">
+        <i className="fa fa-certificate text-info" title="Genesis"></i>
```
msbrogli, line 19: I don't like changing this to a badge. I think we should keep the "GENESIS" text and maybe change it to a badge so others can be added later.

msbrogli (changes requested):

src/components/tx/TxAlerts.js
giovanecosta: I've removed the alert

src/components/tx/TxAlerts.js
giovanecosta: After the introduction of hooks, functional components has become best practice for React. Even the docs encourage it: https://reactjs.org/docs/hooks-faq.html#should-i-use-hooks-classes-or-a-mix-of-both

src/components/tx/TxMarkers.js
giovanecosta: After the introduction of hooks, functional components has become best practice for React. Even the docs encourage it: https://reactjs.org/docs/hooks-faq.html#should-i-use-hooks-classes-or-a-mix-of-both

src/components/tx/TxMarkers.js
giovanecosta: Changed back to text, updated the screenshots

src/components/tx/TxAlerts.js
giovanecosta: yes. removed

src/components/tx/TxMarkers.js
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  const infoMarker = () => {
+    if (!tx.context) {
+      return null;
+    }
+
+    return (
+      <button className="info-hover-wrapper btn btn-link pl-1">
+        <i className="fa fa-certificate text-info" title="Genesis"></i>
```
pedroferreira1, line 29: This icon `title` should come from context, right?

giovanecosta: changed

pedroferreira1 (dismissed):

msbrogli (dismissed):

pedroferreira1 (approved):

src/components/TxData.js
```diff
@@ -399,7 +421,7 @@ class TxData extends React.Component {
       const tokens = this.state.tokens.map((token) => {
         return (
           <div key={token.uid}>
-            <span>{token.name} <strong>({token.symbol})</strong> | {renderTokenUID(token)}</span>
+            <TokenMarkers token={token} /><span>{token.name} <strong>({token.symbol})</strong> | {renderTokenUID(token)}</span>
```
msbrogli, line 424: I feel we should add the same marker in the inputs and outputs.

msbrogli (approved):
Great! Just a minor suggestion and we can merge it.

PR #97, opened 2021-07-29
chore: clear cloudfront cache after deploy

giovanecosta:
closes: https://github.com/HathorNetwork/ops-tools/issues/40

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ testnet_s3_sync:
 	aws s3 sync --delete ./build/ s3://hathor-testnet-foxtrot-public-explorer
 
 .PHONY: testnet_deploy
-testnet_deploy: check_version testnet_s3_sync
+testnet_deploy: check_version testnet_s3_sync clear_cloudfront_cache
 
 .PHONY: mainnet_build
 mainnet_build:
@@ -35,4 +35,8 @@ mainnet_s3_sync:
 	aws s3 sync --delete ./build/ s3://hathor-mainnet-public-explorer
 
 .PHONY: mainnet_deploy
-mainnet_deploy: check_version check_tag mainnet_s3_sync
+mainnet_deploy: check_version check_tag mainnet_s3_sync clear_cloudfront_cache
+
+.PHONY: clear_cloudfront_cache
+clear_cloudfront_cache:
+	aws cloudfront create-invalidation --distribution-id $$CLOUDFRONT_ID --paths "/index.html"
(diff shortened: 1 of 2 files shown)


luislhl (dismissed):

Makefile
```diff
@@ -30,9 +30,13 @@ mainnet_build:
 	export REACT_APP_EXPLORER_SERVICE_BASE_URL=https://explorer-service.hathor.network/; \
 	npm run build
 
-.PHONY: mainnet_s3_sync
+.PHONY: mainnet_s3_sync clear_cloudfront_cache
```
luislhl, line 33: I think this is wrong.

Did it run?

luislhl: We should include `clear_cloudfront_cache` in the `mainnet_deploy`, not here

Makefile
```diff
@@ -30,9 +30,13 @@ mainnet_build:
...
 	aws s3 sync --delete ./build/ s3://hathor-mainnet-public-explorer
 
 .PHONY: mainnet_deploy
-mainnet_deploy: check_version check_tag mainnet_s3_sync
\ No newline at end of file
+mainnet_deploy: check_version check_tag mainnet_s3_sync
+
+.PHONY: clear_cloudfront_cache
+clear_cloudfront_cache:
+	aws cloudfront create-invalidation --distribution-id $$CLOUDFRONT_ID --paths "/index.html"
```
jansegre, line 42: Is paths "/index.html" enough? This wouldn't invalidate "/", right?

luislhl: We had made some tests in the past invalidating just `/index.html` and it worked.

I think `/` is not a path CloudFront is concerned for caching, since it's not a file. `/` is configured to serve the file `/index.html`, so invalidating `/index.html` is enough for `/`.

giovanecosta: Yes, as this is a Single Page App, update only the index.html is enough. JS and CSS files are generated with hash names that will be different.

Makefile
giovanecosta: Yes! My bad. changed

giovanecosta:
Fixed
@luislhl and @jansegre

Makefile
jansegre: OK, I thought CloudFront didn't even know the filenames. So when accessing `/` it serves the `/index.html` file and then it looks on the cache. Fine for me.

jansegre (approved):

luislhl (approved):

PR #99, opened 2021-08-03
fix: we shouldn't show Spent link on inputs

pedroferreira1:
Example tx with problem: https://explorer.testnet.hathor.network/transaction/00005f234469407614bf0abedec8f722bb5e534949ad37650f6077c899741ed7
image

### DIFF
--- a/src/components/TxData.js
+++ b/src/components/TxData.js
@@ -197,13 +197,21 @@ class TxData extends React.Component {
       }
     }
 
-    const renderOutput = (output, idx, addBadge) => {
+    const renderOutputLink = (idx) => {
+      if (idx in this.props.spentOutputs) {
+        return <span> (<Link to={`/transaction/${this.props.spentOutputs[idx]}`}>Spent</Link>)</span>;
+      } else {
+        return null;
+      }
+    }
+
+    const renderOutput = (output, idx, isOutput) => {
       return (
         <div key={idx}>
           <div>{outputValue(output)} {renderOutputToken(output)}</div>
           <div>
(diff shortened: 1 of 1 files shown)


src/components/TxData.js
```diff
@@ -197,13 +197,21 @@ class TxData extends React.Component {
...
-    const renderOutput = (output, idx, addBadge) => {
+    const renderOutputLink = (idx) => {
+      if (idx in this.props.spentOutputs) {
+        return <span> (<Link to={`/transaction/${this.props.spentOutputs[idx]}`}>Spent</Link>)</span>;
+      } else {
+        return null;
+      }
+    }
+
+    const renderOutput = (output, idx, isOutput) => {
```
andreabadesso, line 208: Also change the method name? `renderOutput` with something that is not an output is confusing

andreabadesso (changes requested):

src/components/TxData.js
msbrogli: I agree that it is confusing. But, in this case, we use the same method because we are rendering "the output that the inputs points to". Maybe we should just add a note about it.

msbrogli (approved):

src/components/TxData.js
pedroferreira1: Maybe change to `renderInputOrOutput`?

pedroferreira1: Done

andreabadesso: Looks good

andreabadesso (approved):

PR #105, opened 2021-08-12
Feat/new metadata structure

giovanecosta:


src/components/token/TokenNFTPreview.js
```diff
@@ -49,18 +49,31 @@ const TokenNFTPreview = (props) => {
...
   } else if(nftType === NFT_MEDIA_TYPES.video && fileType) {
     media = (
-      <video controls controlsList="nodownload noremoteplayback" disablePictureInPicture onPlay={onPlayMedia}>
-        <source src={token.meta.nft.file} type={fileType} />
+      <video
+        controls
+        controlsList="nodownload noremoteplayback"
+        disablePictureInPicture
+        onPlay={onPlayMedia}
+        loop={token.meta.data.nft_media.loop}
```
pedroferreira1, line 60: Does it have any problem if `token.meta.data.nft_media.loop` is `undefined`?

pedroferreira1 (approved):

src/components/token/TokenAlerts.js
```diff
@@ -10,7 +10,7 @@ const TokenAlerts = (props) => {
   }, [props.token]);
 
   const bannedAlert = () => {
-    if (!token.banned) {
+    if (!token.meta || !token.meta.data.banned) {
```
andreabadesso, line 13: This expects the token to have `meta.data`, right?

What if the token has the old structure?

src/screens/TokenDetail.js
```diff
@@ -152,14 +152,14 @@ class TokenDetail extends React.Component {
...
-      return this.state.token.meta && this.state.token.meta.nft;
+    const hasNFTMedia = () => {
+      return this.state.token.meta && this.state.token.meta.data.nft_media;
     }
 
     return (
       <div className="content-wrapper flex align-items-center">
         <TokenAlerts token={this.state.token} />
-        { isNFT() ? 
+        { hasNFTMedia() ?
```
andreabadesso, line 162: Why use a ternary when you can just conditionally render inside the method?

Maybe it's personal, but I tend to find ternary operators hard to read and avoidable

andreabadesso (changes requested):

src/components/token/TokenNFTPreview.js
giovanecosta: Nope. It will works just like `false`.

src/components/token/TokenAlerts.js
giovanecosta: It will just evaluate as false, as the old structure doesn't have `meta` property

src/screens/TokenDetail.js
giovanecosta: I agree that in cases where ternary have blocks with multiple lines of html. In this case, we're just deciding what component to render. I think that move it to a function with an `if` and 2 `return` statements is less readable

andreabadesso (approved):

r4mmer:
These changes were made #116

PR #116, opened 2021-09-06
feat: show metadata api permissive

r4mmer:
# Acceptance Criteria

- [x] Use the new metadata api for tokens and transactions
- [x] Show token information as described on the metadata api
- [x] Show transaction information as described on the metadata api
- [x] Presentable, clear and clean display of information from metadata

# Transaction 

image

## Hover:

image

## Transaction from a banned token:

image

# Token

## Scam Token
image

## Verified Token
image

## Others
image


src/components/token/TokenMarkers.js
```diff
@@ -25,16 +25,26 @@ const TokenMarkers = (props) => {
...
 
   const bannedMarker = () => {
-    if (!token.banned) {
+    if (!token.meta || !token.meta.banned) {
       return null;
     }
+    let banIcon = (
+      <i className="fa fa-exclamation-triangle fa-lg text-danger" title="Banned"></i>
+    )
+    if (token.meta && token.meta.reason) {
```
pedroferreira1, line 34: Here you already know that you have `token.meta` because of the `if` above, right?

src/components/token/TokenMarkers.js
```diff
@@ -25,16 +25,26 @@ const TokenMarkers = (props) => {
...
+      banIcon = (
+        <div>
+        <i className="fa fa-exclamation-triangle fa-lg text-danger"></i>
```
pedroferreira1, line 37: Maybe add a tab here inside the `<div>` to improve indentation?

src/components/token/TokenMarkers.js
```diff
@@ -1,5 +1,5 @@
 import React, { useState, useEffect } from 'react';
-import { tokenBannedMessage } from '../../messages';
+// import { tokenBannedMessage } from '../../messages';
```
pedroferreira1, line 2: Why did you remove this? This message is important. We won't show only the metadata reason here. This message is yet to be decided because the wording is been decided with the legal team.

src/messages.js
```diff
@@ -1 +1,2 @@
-export const tokenBannedMessage = ''; // TODO: this message is being disscussed here: https://github.com/HathorNetwork/hathor-explorer/issues/92
\ No newline at end of file
+export const tokenBannedMessage = ''; // TODO: this message is being disscussed here: https://github.com/HathorNetwork/hathor-explorer/issues/92
+export const txTokenBannedMessage = ''; // TODO: this message is being disscussed here: https://github.com/HathorNetwork/hathor-explorer/issues/92
```
pedroferreira1, line 2: What's this message?

src/screens/TransactionDetail.js
```diff
@@ -82,6 +83,32 @@ class TransactionDetail extends React.Component {
...
+   * Get transaction metadata from explorer service
+   */
+  updateTxMetadata = (id) => {
+    metadataApi.getDag(id).then((data) => {
+      if (data) {
+        this.setState((oldState) => {
+          return {
+            transaction: {
+              ...oldState.transaction,
+              meta: data
```
pedroferreira1, line 96: This is getting a bit confusing here. We already have in the state `this.state.meta` which is the metadata that we receive from the full node (the transaction has its own metadata).

I'm not sure about the best name, maybe call `customMeta`? I don't know, what do you think?

src/components/tx/TxAlerts.js
```diff
@@ -0,0 +1,32 @@
...
+
+const TxAlerts = (props) => {
+  const [transaction, setTransaction] = useState(props.transaction);
+
+  useEffect(() => {
+    setTransaction(props.transaction);
+  }, [props.transaction]);
+
+  const bannedAlert = () => {
+    if (!transaction.tokens.find(token => token.meta && token.meta.banned)) {
```
pedroferreira1, line 13: On the `TxData` component you are saving the tokens metadata in the state. Can you get token meta like this?

Or when you do 

```
return metadataApi.getDag(token.uid).then((data) => {
      token.meta = data;
      return token;
```

You are actually mutating the props object?

src/messages.js
r4mmer: This is to show on the `TxAlerts` component whenever a tx contains a token that was banned.

src/components/token/TokenMarkers.js
r4mmer: I'm removing this from token markers because this message belongs on the `TokenAlerts` component

src/components/token/TokenMarkers.js
r4mmer: Done

src/components/token/TokenMarkers.js
r4mmer: Yes, removing the unnecessary check

src/screens/TransactionDetail.js
r4mmer: I was confused as well, the only thing i'm certain is that we don't want to merge metadata from the fullnode and from the explorer-service api. I included on the tx because that's what we do for the token but I think we should change the name

src/components/tx/TxData.js
```diff
@@ -82,7 +85,25 @@ class TxData extends React.Component {
...
+  }
+
+  /**
+   * Fetch token metadata
+   *
+   * @param {Object} token Token object to be updated
+   */
+  getTokenMetadata = (token) => {
+    return metadataApi.getDag(token.uid).then((data) => {
+      token.meta = data;
```
andreabadesso, line 104: `props` coming from redux are not immutable yet, so we should be careful to not mutate them on the components

I suggest refactoring this to something like:

```
...
return metadataApi.getDag(token.uid).then((data) => ({
  ...token,
  meta: data,
}))
...
```

andreabadesso (changes requested):

src/components/tx/TxData.js
r4mmer: Done.

src/components/tx/TxAlerts.js
r4mmer: Changed to a solution proposed by @andreabadesso

andreabadesso (approved):

src/api/metadataApi.js
```diff
@@ -0,0 +1,22 @@
...
+import { requestExplorerServiceV1 } from './axiosInstance';
+
+const metadataApi = {
+  getDag(id) {
+    return requestExplorerServiceV1().get(`metadata/dag`, {params: {id}}).then((res) => {
+      if (id in res.data) {
+        return res.data[id]
+      }
+    }).catch((error) => {
+      // something wrong with request
```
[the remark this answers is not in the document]
pedroferreira1: We should discuss that. We added in the wallet with the user permission. I'm not sure if we should just add this in the explorer without user approval. It would be good but it should be discussed first

src/components/tx/TxAlerts.js
pedroferreira1: That's good but this created a bug in this screen, right? You won't have `token.meta` in the `props.transaction.tokens` array. Or am I missing something?

pedroferreira1: Ok after your last commit

pedroferreira1 (approved):

andreabadesso (approved):

PR #123, opened 2021-09-17
feat: change the way we show amount and token type depending on metadata result

pedroferreira1:
Not NFT screens

image
image
image

NFT screens

image
image
image


src/screens/AddressDetail.js
```diff
@@ -263,13 +270,23 @@ class AddressDetail extends React.Component {
     });
   }
 
+  getSelectedTokenMetadata = (selectedToken) => {
+    metadataApi.getDag(selectedToken).then((data) => {
```
msbrogli, line 274: `getDag`? I feel we should rename it to `getDagMetadata()`.

src/screens/AddressDetail.js
```diff
@@ -263,13 +270,23 @@ class AddressDetail extends React.Component {
...
+      if (data) {
+        this.setState({ selectedTokenMetadata: data });
+      }
+    }).finally(() => {
+      this.setState({ metadataLoaded: true });
```
msbrogli, line 279: What if an error occurs? Should we still set `metadataLoaded` to true?

src/screens/TokenDetail.js
```diff
@@ -94,6 +96,8 @@ class TokenDetail extends React.Component {
           }
         });
       }
+    }).finally(() => {
+      this.setState({ metadataLoaded: true });
```
msbrogli, line 100: Same thing here.

src/utils/helpers.js
```diff
@@ -218,8 +218,25 @@ const helpers = {
...
+   * Render value to integer or decimal
+   *
+   * @param {number} amount Amount to render
+   * @param {boolean} isInteger If it's an integer or decimal
+   *
+   * @return {string} rendered value
+   * @memberof Helpers
+   * @inner
+   */
+  renderValue(amount, isInteger) {
```
msbrogli, line 233: I think this method should be `renderValue(amount, decimals)`. But I understand that it just calls the library which does not support decimals yet, right?

msbrogli (dismissed):

src/screens/AddressDetail.js
pedroferreira1: If the metadata API returns 404 it is loaded as well. But maybe I should capture the error and consider it loaded only if it's 404

src/utils/helpers.js
pedroferreira1: Yes, exactly. Right now the lib supports only 2 decimals or 0 decimals. I guess if we add custom decimal places we will change that (and many other places). I think we should leave that for the future, what do you think?

src/screens/AddressDetail.js
pedroferreira1: Makes sense. Changed that

src/screens/AddressDetail.js
pedroferreira1: Done

src/screens/TokenDetail.js
pedroferreira1: Done

src/utils/helpers.js
msbrogli: I agree.

andreabadesso (approved):

msbrogli (approved):

Issue #128, opened 2021-09-30
network page goes to undefined on node deploy


r4mmer:
# Description

During deploy of a node the  `/network` page the explorer goes to `/network/undefined`.
This is probably because the explorer-service is returning an empty list of available peers even though there are some available peers.

# Acceptance criteria

- Have the explorer return the available peers accurately.
- Have something to show if there are no available peers.
- Retry the api to load peers until there is a peer available.

PR #135, opened 2021-10-11
fix: network page peer goes undefined during deploy

r4mmer:
#128 

This will not resolve the problem of the daemon not registering the keys, but it will prevent the peer to be `undefined` when the list of peers goes empty.

### DIFF
--- a/src/components/Network.js
+++ b/src/components/Network.js
@@ -48,6 +48,9 @@ class Network extends React.Component {
     const { match: { params } } = this.props;
 
     networkApi.getPeerList().then((peers) => {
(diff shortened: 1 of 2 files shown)


src/components/Network.js
```diff
@@ -48,14 +48,18 @@ class Network extends React.Component {
     const { match: { params } } = this.props;
 
     networkApi.getPeerList().then((peers) => {
+      if (peers.error === undefined && peers.length === 0) {
+        throw new Error("No peers present.");
```
luislhl, line 52: What does the user see in case we throw this error, while it's not retried?

PR #141, opened 2021-11-30
feat: decode output script data in the tx detail screen

pedroferreira1:
### Acceptance criteria

- Output script data shouldn't show 'unable to decode' anymore, so any NFT creation tx will show the decoded script data.

### TODO

- Bump wallet lib version after release

Old screen
image

New screen
image

### DIFF
--- a/src/components/tx/TxData.js
+++ b/src/components/tx/TxData.js
@@ -279,12 +279,29 @@ class TxData extends React.Component {
           return renderNanoContractMatchValues(output.decoded);
         default:
           let script = output.script;
+          // Try to parse as script data
+          try {
+            // The output script is decoded to base64 in the full node
+            // before returning as response to the explorer in the API
+            // and the lib expects a buffer (bytes)
+            // In the future we must receive from the full node
+            // the decoded.type as script data but this still needs
+            // some refactor there that won't happen soon
+            const buff = new Buffer.from(script, 'base64');
+            const parsedData = hathorLib.scriptsUtils.parseScriptData(buff);
+            return renderDataScript(parsedData.data);
+          } catch {}
+
           try {
             script = atob(output.script)
           } catch {}
 
           return `Unable to decode script: ${script.trim()}`;
       }
+    }
+
+    const renderDataScript = (data) => {
+      return `${data} [Data Script]`;
     }
 
     const renderP2PKHorMultiSig = (decoded) => {
--- a/src/components/tx/TxData.js
+++ b/src/components/tx/TxData.js
@@ -290,7 +290,13 @@ class TxData extends React.Component {
             const buff = new Buffer.from(script, 'base64');
             const parsedData = hathorLib.scriptsUtils.parseScriptData(buff);
             return renderDataScript(parsedData.data);
-          } catch {}
+          } catch(e) {
+            if (!(e instanceof hathorLib.errors.ParseScriptError)) {
+              // Parse script error is the expected error in case the output script
+              // is not a script data. If we get another error here, we should at least log it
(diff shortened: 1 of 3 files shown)


r4mmer (dismissed):

andreabadesso (dismissed):

src/components/tx/TxData.js
```diff
@@ -279,6 +279,19 @@ class TxData extends React.Component {
...
+            // The output script is decoded to base64 in the full node
+            // before returning as response to the explorer in the API
+            // and the lib expects a buffer (bytes)
+            // In the future we must receive from the full node
+            // the decoded.type as script data but this still needs
+            // some refactor there that won't happen soon
+            const buff = new Buffer.from(script, 'base64');
+            const parsedData = hathorLib.scriptsUtils.parseScriptData(buff);
+            return renderDataScript(parsedData.data);
+          } catch {}
```
msbrogli, line 293: This exception handler might mask bugs in the parsing script. Shouldn't we capture a specific type of error here?

src/components/tx/TxData.js
```diff
@@ -287,6 +300,10 @@ class TxData extends React.Component {
       }
     }
 
+    const renderDataScript = (data) => {
+      return `${data} [Data Script]`;
```
msbrogli, line 304: Maybe just `[Data]`.

msbrogli (changes requested):

src/components/tx/TxData.js
pedroferreira1: We receive the output script from the full node in base64.

The old behaviour of this parse method was to try to decode the base64 and show 'Unable to decode: X' and it would show X as the decoded base64 or whatever the full node has sent to us as output script. I tried to keep the same idea, if we can't parse the script we will show 'Unable to decode: X' where X is what we received from the full node.

I understand that the best approach is usually to capture the specific error but I guess the idea here is to prevent any exception in the explorer and show as 'Unable to decode', as the previous code used to do.

src/components/tx/TxData.js
pedroferreira1: I agree it's better. Changed here: https://github.com/HathorNetwork/hathor-explorer/pull/141/commits/61751ea06f66ec5e82c084941af18b7fbca506a9

src/components/tx/TxData.js
msbrogli: I understand. But a bug is different from an unknown script. We should at least log the error in case of a bug.

pedroferreira1: Makes sense. Added the handling and log here https://github.com/HathorNetwork/hathor-explorer/pull/141/commits/66d891268e9f16682e776c0e7b8e116ff90be3db

src/components/tx/TxData.js
```diff
@@ -279,6 +279,25 @@ class TxData extends React.Component {
...
+            // the decoded.type as script data but this still needs
+            // some refactor there that won't happen soon
+            const buff = new Buffer.from(script, 'base64');
+            const parsedData = hathorLib.scriptsUtils.parseScriptData(buff);
+            return renderDataScript(parsedData.data);
+          } catch(e) {
+            if (!(e instanceof hathorLib.errors.ParseScriptError)) {
+              // Parse script error is the expected error in case the output script
+              // is not a script data. If we get another error here, we should at least log it
+              console.log('Error parsing script data', e);
```
msbrogli, line 297: I'd say "Unexpected error".

msbrogli (approved):

andreabadesso (dismissed):

andreabadesso (approved):

PR #154, opened 2021-12-24
feat: remove token list screen and peer count in stats

pedroferreira1:
### Acceptance Criteria

- We shouldn't see the peer count in the Statistics screen because this count is not 100% reliable.
- We should remove the tokens list screen.


r4mmer (approved):
Should we also remove the token proxy on explorer service? It was going to be used only by the tokens screen.

andreabadesso (approved):

r4mmer (dismissed):

r4mmer (approved):

PR #155, opened 2021-12-27
feat: use explorer service node api proxy endpoints

r4mmer:
# Acceptance criteria 

Use proxy apis from explorer-service instead of requesting the full-node directly.

depends on 3 apis implemented on HathorNetwork/hathor-explorer-service#109


pedroferreira1 (approved):

andreabadesso (approved):
@r4mmer 
Is there some specific reason for not using async/await on the explorer? Or we are just being consistent with the current code?

r4mmer:
> Is there some specific reason for not using async/await on the explorer? Or we are just being consistent with the current code?

I'm not aware of any impediments, this was done for consistency, but we could make a refactor to start using more async where needed.

pedroferreira1 (approved):

PR #161, opened 2022-01-26
feat(statistics): add orphan blocks and remove block info (#159)

lucas3003:
### Motivation

To give a better user experience on statistics page, we are removing "block" and adding "orphan blocks" information.

### Acceptance criteria

The explorer must show the statistics page as this screenshot:

Screenshot from 2022-01-26 10-57-21


pedroferreira1 (dismissed):

r4mmer (approved):

src/screens/Dashboard.js
```diff
@@ -39,8 +39,8 @@ class Dashboard extends React.Component {
 
     return (
       <div className="content-wrapper">
-        <p><strong>Blocks: </strong>{helpers.renderValue(blocks, true)}</p>
         <p><strong>Height of the best chain: </strong>{helpers.renderValue(height, true)}</p>
+        <p><strong>Orphan blocks: </strong>{helpers.renderValue(orphanBlocks, true)}</p>
```
msbrogli, line 43: As briefly discussed in our daily meeting, I feel that we shouldn't add the orphan blocks here. My point is that orphan blocks are not synced among peers and this number must be different for each peer. Besides that, this information is not useful except for miners. In this case, miners will keep track of the orphan blocks so they can improve their mining strategy to reduce it as much as possible.

The explorer service might keep track of the number of orphan blocks. It might be useful for us to understand how efficient the mining is, and maybe get in touch with miners to help them improve their operations.

@pedroferreira1 @r4mmer @lucas3003 What do you think?

msbrogli (changes requested):

src/screens/Dashboard.js
pedroferreira1: I agree. I talked with @lucas3003 yesterday that I misunderstood some discussions on Slack and created the issue not how it should be.

lucas3003: I agree. I will remove "Blocks" line and rename "Height of the best chain" to "Blocks (best height)".

lucas3003: Also, I will send that in a new PR. I mistakenly created the branch on top of master instead of dev.

lucas3003: Just sent the new PR: https://github.com/HathorNetwork/hathor-explorer/pull/164

Issue #158, opened 2022-01-13
Transaction detail screen: graphviz graphs should be loaded only when requested by user


pedroferreira1:
We currently have two graphs (funds and verification neighbors) inside a transaction detail screen. We should load and show them only if the users clicks a button or a link "Load verification neighbors", not always.

PR #165, opened 2022-01-28
feat(transactions): load neighbors graphs only when requested by user (#158)

lucas3003:
### Motivation

Today, when the user loads the transaction page, it automatically loads the neighbors graphs (verification and funds). If the user opens many transactions, the interface starts to slow down because of these automatically loaded graphs.

To solve this problem, a lazy loading mechanism was introduced. Now, the graph will only be loaded when user requests.

### Acceptance Criteria

User must be able to visualize neighbors graphs when clicking on "Click to show" and hiding it clicking on "Click to hide", as shown on this video:

Obs. A sleep of 5 seconds was intentionally placed to simulate a scenario where the full-node is taking too long to respond:

https://user-images.githubusercontent.com/3287486/151799933-a3af6eeb-df32-428d-acdb-cf4c9a6f3513.mov

### DIFF
--- a/src/components/tx/TxData.js
+++ b/src/components/tx/TxData.js
@@ -137,12 +137,11 @@ class TxData extends React.Component {
   calculateNeighborsGraph = async (graphType) => {
     const viz = new Viz({ Module, render });
 
-    await graphvizApi.dotNeighbors(this.props.transaction.hash, graphType).then(response => {
-      viz.renderSVGElement(response).then((element) => {
-          element.id = `graph-${graphType}`;
-          document.getElementById(`graph-${graphType}`).appendChild(element);
-      });
-    });
+    const graphvizResponse = await graphvizApi.dotNeighbors(this.props.transaction.hash, graphType);
+    const element = await viz.renderSVGElement(graphvizResponse);
+
+    element.id = `graph-${graphType}`;
+    document.getElementById(`graph-${graphType}`).appendChild(element);
   }
 
   /**
@@ -159,12 +158,15 @@ class TxData extends React.Component {
 
     this.setState({ graphs });
 
+    // Check if graph needs to be calculated before showing
     if(!graphs[index].calculatedNeighbors && !graphs[index].graphLoading) {
       graphs[index].graphLoading = true;
       this.setState({ graphs });
 
+      // Make the necessary requests to calculate the graph
       await this.calculateNeighborsGraph(graphs[index].name);
 
+      // Update graph status
       graphs[index].calculatedNeighbors = true;
       graphs[index].graphLoading = false;
       this.setState({ graphs });
--- a/src/components/tx/TxData.js
+++ b/src/components/tx/TxData.js
@@ -34,40 +34,33 @@ class TxData extends React.Component {
    * children {boolean} if should show children (default is hidden but user can show with a click)
    * tokens {Array} tokens contained in this transaction
    * metadataLoaded {boolean} true when all token metadatas are loaded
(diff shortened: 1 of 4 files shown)


src/components/tx/TxData.js
```diff
@@ -409,10 +448,12 @@ class TxData extends React.Component {
...
     }
 
-    const renderGraph = (label, type) => {
+    const renderGraph = (label, type, toggleFunction, currentState) => {
       return (
-        <div className="mt-3 graph-div" id={`graph-${type}`} key={`graph-${type}-${this.props.transaction.hash}`}>
+        <div  id={`graph-neighbors`} key={`graph-${type}-${this.props.transaction.hash}`}>
           <label className="graph-label">{label}:</label>
+          <a href="true" className="ml-1" onClick={(e) => toggleFunction(e)}>{currentState ? 'Click to hide' : 'Click to show'}</a>
+          <div className={currentState ? undefined : 'hidden'} id={`graph-${type}`}></div>
```
pedroferreira1, line 456: If the full node is loaded with requests it may take some time to load this graph and the user will see only an empty box. Maybe we could create another state `fundsGraphLoaded` (the same for verification) and inside this `div` we add a `{!fundsGraphLoaded && 'Loading graph...'}` so the user understands that something is loading. We can even add a Loading that is already in use in the project (`ReactLoading`)

What do you think?

lucas3003: Great idea, I think this really improves the UX. Currently, if nothing happens for a while, the user will think the action did not work. He will try to reload the page, click again and generate more load on full-node.

pedroferreira1 (approved):

src/components/tx/TxData.js
```diff
@@ -409,10 +461,15 @@ class TxData extends React.Component {
       )
     }
 
-    const renderGraph = (label, type) => {
+    const renderGraph = (graphIndex) => {
       return (
-        <div className="mt-3 graph-div" id={`graph-${type}`} key={`graph-${type}-${this.props.transaction.hash}`}>
-          <label className="graph-label">{label}:</label>
+        <div className="d-flex flex-column flex-lg-row align-items-start mb-3 common-div bordered-wrapper w-100">
+          <div className="mt-3 graph-div" id={`graph-neighbors`} key={`graph-${this.state.graphs[graphIndex].name}-${this.props.transaction.hash}`}>
```
r4mmer, line 467: Is this id used? If so we should make it something like `graph-neighbors-${graphIndex}` (or name) to differentiate.

src/components/tx/TxData.js
```diff
@@ -121,6 +129,50 @@ class TxData extends React.Component {
...
+   *
+   * @param {string} graphType
+   */
+  calculateNeighborsGraph = async (graphType) => {
+    const viz = new Viz({ Module, render });
+
+    const graphvizResponse = await graphvizApi.dotNeighbors(this.props.transaction.hash, graphType);
+    const element = await viz.renderSVGElement(graphvizResponse);
+
+    element.id = `graph-${graphType}`;
```
r4mmer, line 143: I know we won't recalculate this, but we shouldn't add an element with the same id as the parent. Maybe use `graph-${graphType}-data`.

r4mmer (changes requested):

src/components/tx/TxData.js
lucas3003: It is not being used. I will remove it.

r4mmer (approved):

pedroferreira1 (approved):

PR #180, opened 2022-04-08
Retrieve tokens data from explorer-service

lucas3003:
### Acceptance Criteria

User must be able to navigate through Tokens page, searching for specific tokens and sorting by fields available on table header.

Consideration:

- After a table header is clicked to sort data, the table will go back to the first page.
- Sort by last column (Type) was not implemented, as only two values are possible (Custom Token or NFT). 

### Video

https://user-images.githubusercontent.com/3287486/163101019-2012e3c9-d35b-4bf5-9ed6-8b0bbd6735e4.mov

### DIFF
--- a/src/components/token/TokenSearchField.js
+++ b/src/components/token/TokenSearchField.js
@@ -6,7 +6,7 @@ class TokenSearchField extends React.Component {
         return (
             <div className="d-flex flex-row align-items-center navigation-search-token">
                 <div className="d-flex flex-row align-items-center col-12">
-                    <input className="form-control mr-2" type="search" value={this.props.searchText} onChange={this.props.onSearchTextChanged} placeholder="Search UID, name, symbol, or type" aria-label="Search" ref="tokenSearch" />
+                    <input className="form-control mr-2" type="search" value={this.props.searchText} onKeyPress={this.props.onSearchTextKeyPressed} onChange={this.props.onSearchTextChanged} placeholder="Search UID, name, symbol, or type" aria-label="Search" ref="tokenSearch" />
                     <i className="fa fa-search pointer" onClick={(e) => this.props.onSearchButtonClicked(e)}></i>
                 </div>
             </div>
--- a/src/components/token/TokenSearchField.js
+++ b/src/components/token/TokenSearchField.js
@@ -1,13 +1,17 @@
 import React from 'react'
 import PropTypes from 'prop-types';
+import Loading from '../Loading';
 
 class TokenSearchField extends React.Component {
     render() {
         return (
             <div className="d-flex flex-row align-items-center navigation-search-token">
                 <div className="d-flex flex-row align-items-center col-12">
-                    <input className="form-control mr-2" type="search" value={this.props.searchText} onKeyPress={this.props.onSearchTextKeyPressed} onChange={this.props.onSearchTextChanged} placeholder="Search UID, name, symbol, or type" aria-label="Search" ref="tokenSearch" />
-                    <i className="fa fa-search pointer" onClick={(e) => this.props.onSearchButtonClicked(e)}></i>
+                    <input className="form-control mr-2 search-input" type="search" value={this.props.searchText} onKeyUp={this.props.onSearchTextKeyUp} onChange={this.props.onSearchTextChanged} placeholder="Search UID, name, symbol, or type" aria-label="Search" ref="tokenSearch" />
+                    {(this.props.isSearchLoading && !this.props.loading) ?
+                        <Loading width={25} height={25} delay={0} useLoadingWrapper={false} showSlowLoadMessage={false} /> :
+                        <i className="fa fa-search pointer" onClick={(e) => this.props.onSearchButtonClicked(e)} />
+                    }
                 </div>
             </div>
         )
--- a/src/components/token/Tokens.js
+++ b/src/components/token/Tokens.js
@@ -10,7 +10,7 @@ import PropTypes from 'prop-types';
 import TokensTable from './TokensTable';
 import TokenSearchField from './TokenSearchField';
 import tokensApi from '../../api/tokensApi';
-import { get, forEach, isNil, last, find, isEmpty } from 'lodash';
+import { get, last, find, isEmpty } from 'lodash';
 
 import {
     shouldRenderCustomTokens
@@ -23,6 +23,19 @@ class Tokens extends React.Component {
     constructor(props) {
         super(props);
 
+        /**
+         * tokens: List of tokens currently being rendered
+         * hasAfter: Indicates if a next page exists
+         * hasBefore: Indicates if a previous page exists
+         * searchText: Input text written by user
+         * sortBy: Which field to sort (uid, name, symbol)
+         * order: If sorted field must be ordered asc or desc
+         * page: Current page. Used to know if there is a previous page
+         * pageSearchAfter: searchAfter param needed to get the next page
+         * loading: Initial loading, when user clicks on the Tokens navigation item
+         * isSearchLoading: Indicates if search results are being retrieved from explorer-service
+         * calculatingPage: Indicates if next page is being retrieved from explorer-service
+         */
         this.state = {
             tokens: [],
             hasAfter: false,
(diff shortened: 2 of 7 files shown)


src/components/token/Tokens.js
```diff
@@ -24,18 +27,101 @@ class Tokens extends React.Component {
...
+                searchAfter: []
+            }
+        ]})
+    }
+
+    getTokens = async (searchAfter) => {
+        const tokensRequest = await tokensApi.getList(this.state.searchText, this.state.sortBy, this.state.order, searchAfter)
+        let tokens = get(tokensRequest, 'data', {hits:[], 'has_next': false})
+        forEach(tokens.hits, (token) => {
+            if(isNil(token.uid)) {
```
pedroferreira1, line 59: The backend is sending sometimes `uid` and sometimes `id`?

src/components/token/Tokens.js
```diff
@@ -24,18 +27,101 @@ class Tokens extends React.Component {
...
+                token.uid = token.id
+            }
+            token.type = token.nft ? 'NFT' : 'Custom Token'
```
pedroferreira1, line 62: If we remove this loop, we can do this in the `<td>` itself

src/components/token/TokenSearchField.js
```diff
@@ -6,7 +6,7 @@ class TokenSearchField extends React.Component {
         return (
             <div className="d-flex flex-row align-items-center navigation-search-token">
                 <div className="d-flex flex-row align-items-center col-12">
-                    <input className="form-control mr-2" type="search" value={this.props.searchText} onChange={this.props.onSearchTextChanged} placeholder="Search UID, name, symbol, or type" aria-label="Search" ref="tokenSearch" />
+                    <input className="form-control mr-2" type="search" value={this.props.searchText} onKeyPress={this.props.onSearchTextKeyPressed} onChange={this.props.onSearchTextChanged} placeholder="Search UID, name, symbol, or type" aria-label="Search" ref="tokenSearch" />
```
pedroferreira1, line 9: If the user keeps 'enter' pressed, it will trigger this twice (or more). Should we change to keyup (or keydown) event?

src/components/token/Tokens.js
```diff
@@ -24,18 +27,101 @@ class Tokens extends React.Component {
...
+        this.setState({pageSearchAfter: [
+            {
+                page: 1,
+                searchAfter: []
+            }
+        ]})
+    }
+
+    getTokens = async (searchAfter) => {
+        const tokensRequest = await tokensApi.getList(this.state.searchText, this.state.sortBy, this.state.order, searchAfter)
```
pedroferreira1, line 56: In a UX perspective, I believe we should start the loading icon every time we execute a request. The video seems good because the internet connection was fast but we should protect ourselves from a slow request and I think giving the user a UI feedback is good enough.

src/components/token/Tokens.js
```diff
@@ -9,6 +9,9 @@ import React from 'react';
 import PropTypes from 'prop-types';
```
pedroferreira1, line 9: Comment unrelated to the line.

We don't have a linter and we should add it to keep the code consistent and a standard. Just some comments that I've seen from the review and I think could be refactored:

1. single space after `if` statement, e.g. `if (condition)`.
2. `;` in the end of the line.
3. Single space after starting object and before closing it, e.g. `this.setState({ loading: false })`.
4. Single space around operators, e.g. `const nextPage = this.state.page + 1`

We should also add docstring to all methods in this file with a short explanation of what the method does.

src/components/token/Tokens.js
```diff
@@ -24,18 +27,101 @@ class Tokens extends React.Component {
             tokens: [],
             hasAfter: false,
             hasBefore: false,
-            searchText: ""
+            searchText: "",
```
pedroferreira1, line 30: We should add docstring with each state with brief description and type

src/components/token/Tokens.js
```diff
@@ -24,18 +27,101 @@ class Tokens extends React.Component {
...
-    onSearchButtonClicked = () => {
-        // TODO: Implement behavior when search button is clicked
+    componentDidMount = async() => {
+        //"Click" on search to make the first query
+        await this.onSearchButtonClicked();
+        this.setState({loading: false})
+    }
+
+    resetPageSearchAfter = () => {
+        this.setState({pageSearchAfter: [
```
pedroferreira1, line 47: What's the expected format of elements in `pageSearchAfter` array?

src/components/token/Tokens.js
```diff
@@ -24,18 +27,101 @@ class Tokens extends React.Component {
...
+        }
+    }
+
+    nextPageClicked = async (event) => {
+        const nextPage = this.state.page+1
+        let searchAfter = get(find(this.state.pageSearchAfter, {page: nextPage}), 'searchAfter', [])
+
+        //Calculate searchAfter of next page if not already calculated
+        if(isEmpty(searchAfter)) {
+            const lastCurrentTokenSort = get(last(this.state.tokens), 'sort', []);
```
pedroferreira1, line 92: What's the expected format of elements in `tokens` array?

src/components/token/Tokens.js
```diff
@@ -24,18 +27,101 @@ class Tokens extends React.Component {
...
+            searchAfter = lastCurrentTokenSort
+        }
+
+        const tokens = await this.getTokens(searchAfter);
+        this.setState({tokens: tokens.hits, hasAfter: tokens.has_next, hasBefore: true, page: nextPage})
+    }
+
+    previousPageClicked = async(event) => {
+        const previousPage = this.state.page-1;
+        const searchAfter = get(find(this.state.pageSearchAfter, {page: previousPage}), 'searchAfter', [])
```
pedroferreira1, line 109: What if we don't have this `searchAfter` in the state?

I understand that with "normal" use, you start on page 1, then to go back you have to have this element on state but we currently have a task to add the search params in the querystring so we can easily share links. In that case, you can start your component on page 2. In that case, how do you go back?

pedroferreira1 (changes requested):

src/components/token/Tokens.js
lucas3003: Backend is sending as `id`. And as we are using `id` and `uid` as interchangeable terms in this context, I added this verification to make sure we get the value independently of what we receive from backend.

src/components/token/Tokens.js
lucas3003: I will send this change in the next commit

src/components/token/TokenSearchField.js
lucas3003: I will change it to keyup.

src/components/token/Tokens.js
lucas3003: pageSearchAfter calculates the searchAfter param that needs to be passed to explorer-service in order to get the next/previous page. We use this array to store already-calculated values, so we do not need to recalculate them if user is requesting an already-navigated page.

src/components/token/Tokens.js
lucas3003: Changed it. I've added loading icons. I will update on the next commit and also change the video on PR description

src/components/token/Tokens.js
lucas3003: Each `tokens` element must be an object with `uid,` `name,` `symbol,`  `nft`, and `sort`.

`sort` is an array with two strings. The value is given by ElasticSearch and it is passed back when we want to change page.

src/components/token/Tokens.js
lucas3003: We do have this limitation. We might need to limit querystring to `searchText`, `orderBy`, and `sortBy`

src/components/token/Tokens.js
r4mmer: I understand how `sort` and `searchAfter` work on the ES query context, but should we rename on our api return so it doesn't get confusing?

src/components/token/Tokens.js
```diff
@@ -20,20 +23,155 @@ class Tokens extends React.Component {
...
+        return tokens;
+    }
+
+    /**
+     * Process events when user clicks on search button
+     */
+    onSearchButtonClicked = async () => {
+        this.setState({ isSearchLoading: true });
+        //When search button is clicked, results return to the first page
+        this.resetPageSearchAfter();
```
r4mmer, line 94: We only use this method here, since it only is a `setState`, why not incorporate it on the `setState` below?

src/components/token/Tokens.js
```diff
@@ -20,20 +23,155 @@ class Tokens extends React.Component {
...
+         * isSearchLoading: Indicates if search results are being retrieved from explorer-service
+         * calculatingPage: Indicates if next page is being retrieved from explorer-service
+         */
         this.state = {
             tokens: [],
             hasAfter: false,
             hasBefore: false,
-            searchText: ""
+            searchText: "",
+            sortBy: "uid",
```
r4mmer, line 44: Maybe sort by name, symbol or timestamp by default

src/components/token/Tokens.js
```diff
@@ -20,20 +23,155 @@ class Tokens extends React.Component {
...
+     * Process table header click. This indicates that user wants data to be sorted by a determined field
+     *
+     * @param {*} event
+     * @param {*} header
+     */
+    tableHeaderClicked = async (event, header) => {
+        if (header === this.state.sortBy) {
+            await this.setState({ order: this.state.order === "asc" ? "desc" : "asc" });
+        } else {
+            await this.setState({ sortBy: header });
```
r4mmer, line 171: Maybe set `order` to `asc` here

r4mmer (changes requested):

src/components/token/Tokens.js
pedroferreira1: I think it's fine for now and we can discuss that in the queryparam PR

src/components/token/Tokens.js
pedroferreira1: I feel this could be added to the state docstring so it's clear in the code

src/components/token/Tokens.js
pedroferreira1: I think all those fields and explanation should be added to the state docstring. For me this would be enough to be honest.

pedroferreira1 (approved):

src/components/token/Tokens.js
lucas3003: We will sort by timestamp, but we will need to change Logstash, explorer-service and explorer. This will be too much for this PR and I will send a new one with this change. Does that sound good to you?

src/components/token/Tokens.js
lucas3003: Good catch. I will change on the next commit

src/components/token/Tokens.js
lucas3003: Will change on next commit

src/components/token/Tokens.js
lucas3003: I've added the state docstring

src/components/token/Tokens.js
lucas3003: I've updated the state docstring with this explanation.

pedroferreira1: >         * tokens: List of tokens currently being rendered

This should be improved in my opinion. You should write what you said in the comment so nobody will have questions anymore.

> Each tokens element must be an object with uid, name, symbol, nft, and sort.
> sort is an array with two strings. The value is given by ElasticSearch and it is passed back when we want to change page.

pedroferreira1 (dismissed):

pedroferreira1 (approved):

r4mmer (approved):

PR #184, opened 2022-04-20
feat: add querystring to tokens page

lucas3003:
### Acceptance Criteria

Every query made on tokens page must generate a query string (Even the query that is made when user first access the page).

Example: ```/tokens?searchText=tes1&sortBy=uid&order=asc```

### DIFF
--- a/src/components/token/Tokens.js
+++ b/src/components/token/Tokens.js
@@ -116,6 +116,8 @@ class Tokens extends React.Component {
                 }
             ]
         });
+
+        // This is ultimately called when search text, sort, or sort order changes
         this.updateURL();
     }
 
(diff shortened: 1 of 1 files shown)


r4mmer (approved):

r4mmer:
I approved the code, but maybe add a video of the UI

src/components/token/Tokens.js
```diff
@@ -97,6 +116,7 @@ class Tokens extends React.Component {
                 }
             ]
         });
+        this.updateURL();
```
pedroferreira1, line 119: I would just add a comment here that this will update the URL when the search text, sort, or sort order changes. I had to look at the `tableHeaderClicked` method to understand it also calls `onSearchButtonClicked`.

pedroferreira1 (approved):

pedroferreira1 (approved):

r4mmer (approved):

PR #197, opened 2022-05-30
feat: added token balances page and integration with elasticsearch

andreabadesso:


src/components/SortableTable.js
```diff
@@ -0,0 +1,122 @@
...
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+import PropTypes from 'prop-types';
+import Loading from './Loading';
+import ErrorMessageWithIcon from './error/ErrorMessageWithIcon';
+
+class SortableTable extends React.Component {
```
lucas3003, line 13: Good job generalizing the table

src/components/token/TokenBalances.js
```diff
@@ -0,0 +1,327 @@
...
+    });
+
+    return get(tokenBalanceInformationRequest, 'data', {
+      transactions: 0,
+      addresses: 0,
+    });
+  }
+
+  /**
+   * Calls elastic search with state data, sets loading and URL information
```
lucas3003, line 123: I think we need to explain that ElasticSearch is not called directly, but with explorer-service as gateway

lucas3003 (dismissed):

src/components/token/TokenBalances.js
```diff
@@ -0,0 +1,327 @@
...
+    this.updateURL();
+  }
+
+  /**
+   * Update the URL, so user can share the results of a search
+   */
+  updateURL = () => {
+    const newURL = this.pagination.setURLParameters({
+      sortBy: this.state.sortBy,
+      order: this.state.order,
```
luislhl, line 154: Shouldn't we add the search term as a URL parameter as well, to allow sharing the performed search with someone else?

src/components/token/TokenBalances.js
```diff
@@ -0,0 +1,327 @@
...
+     * page: Current page. Used to know if there is a previous page
+     * pageSearchAfter: Calculates the searchAfter param that needs to be passed to explorer-service in order to get the next/previous page.
+     *                  We use this array to store already-calculated values,
+     *                  so we do not need to recalculate them if user is requesting an already-navigated page.
+     * loading: Initial loading, when user clicks on the Tokens navigation item
+     * isSearchLoading: Indicates if search results are being retrieved from explorer-service
+     * calculatingPage: Indicates if next page is being retrieved from explorer-service
+     * error: Indicates if an unexpected error happened when calling the explorer-service
+     * tokenBalanceInformationError: Indicates if an unexpected error happened when calling the token balance information service 
+     * maintenanceMode: Indicates if explorer-service or its downstream services are experiencing problems. If so, maintenance mode is enabled as
```
luislhl, line 52: Maybe this description is slightly misleading.

By reading this, I could think that there is really a circuit breaking mechanism implemented that will enable the `maintenanceMode` automatically if it detects problems.

But in reality we just have the flag in Unleash which is activated manually, right?

src/components/token/TokenBalances.js
```diff
@@ -0,0 +1,327 @@
...
+    if (this.state.maintenanceMode) {
+      return <ErrorMessageWithIcon message='This feature is under maintenance. Please try again after some time' />;
+    }
+
+    const renderSearchField = () => {
+      return <TokenAutoCompleteField onTokenSelected={this.onTokenSelected.bind(this)}/>;
+    };
+
+    const renderTokensTable = () => {
+      if (this.state.maintenanceMode) {
```
luislhl, line 261: Why to we need this? Won't we already return the `<ErrorMessageWithIcon />` above in case this is enabled?

src/components/token/TokenBalances.js
```diff
@@ -0,0 +1,327 @@
...
+      hasAfter: true,
+      hasBefore: previousPage === 1 ? false : true,
+      page: previousPage,
+      calculatingPage: false,
+    });
+  }
+
+  onTokenSelected = async (token) => {
+    if (!token) {
+      await this.setState({
```
luislhl, line 220: Does it make sense to await `this.setState`? I think it doesn't return a Promise to be awaited

src/components/token/TokenBalances.js
```diff
@@ -0,0 +1,327 @@
...
+   * Call explorer-service to get list of token balances for a given tokenId
+   *
+   * @param {*} searchAfter Parameter needed by ElasticSearch for pagination purposes
+   * @returns tokens
+   */
+  getTokenBalances = async (searchAfter) => {
+    const tokenBalancesRequest = await tokensApi.getBalances(this.state.tokenId, this.state.sortBy, this.state.order, searchAfter);
+    this.setState({
+      error: get(tokenBalancesRequest, 'error', false),
+    });
```
luislhl, line 106: I think setting this every time will trigger unnecessary re-renders in the component.

> The default behavior is to re-render on every state change, and in the vast majority of cases you should rely on the default behavior.

https://reactjs.org/docs/react-component.html#shouldcomponentupdate

Unless we turn this component into a `PureComponent`, which would perform a shallow comparison in the state before triggering a re-render.

But maybe we should just set the state in case there is actually an error.

src/components/token/TokenBalances.js
```diff
@@ -0,0 +1,327 @@
...
+
+  componentDidMount = async () => {
+    if (this.state.maintenanceMode) {
+      return;
+    }
+
+    // 'Click' on search to make the first query
+    const queryParams = this.pagination.obtainQueryParams();
+
+    await this.setState({
```
luislhl, line 83: I think `this.setState` is not awaitable, as I detailed in another comment.

src/components/token/TokenBalances.js
```diff
@@ -0,0 +1,327 @@
...
+    if (!token) {
+      await this.setState({
+        tokenId: '00'
+      });
+
+      this.performSearch();
+      return;
+    }
+
+    await this.setState({
```
luislhl, line 228: I think `this.setState` is not awaitable, as I detailed in another comment.

src/components/token/TokenBalances.js
```diff
@@ -0,0 +1,327 @@
...
+   * Process table header click. This indicates that user wants data to be sorted by a determined field
+   *
+   * @param {*} event
+   * @param {*} header
+   */
+  tableHeaderClicked = async (_event, header) => {
+    if (header === this.state.sortBy) {
+      await this.setState({ order: this.state.order === 'asc' ? 'desc' : 'asc' });
+    } else {
+      await this.setState({ sortBy: header, order: 'asc' });
```
luislhl, line 245: I think `this.setState` is not awaitable, as I detailed in another comment.

src/components/token/TokenBalances.js
```diff
@@ -0,0 +1,327 @@
...
+
+  /**
+   * Process table header click. This indicates that user wants data to be sorted by a determined field
+   *
+   * @param {*} event
+   * @param {*} header
+   */
+  tableHeaderClicked = async (_event, header) => {
+    if (header === this.state.sortBy) {
+      await this.setState({ order: this.state.order === 'asc' ? 'desc' : 'asc' });
```
luislhl, line 243: I think `this.setState` is not awaitable, as I detailed in another comment.

src/components/token/TokenBalances.js
andreabadesso: It is not well documented, but does returns a promise 

This code has a race condition without it

luislhl: Amd because of this, I think we are in danger of falling in this pitfall here as well:

> setState() does not always immediately update the component. It may batch or defer the update until later. This makes reading this.state right after calling setState() a potential pitfall.

https://reactjs.org/docs/react-component.html#setstate

We are setting the `sortBy` and `order` in the state, and we will need to read them inside `this.performSearch` which is called immediately after.

Maybe we should call `this.pagination.obtainQueryParams();` in the constructor instead and set the initial state there. (I think the query params are available at construction time)

src/api/tokensApi.js
```diff
@@ -8,28 +8,86 @@
 import requestExplorerServiceV1 from './axiosInstance';
 import { get } from 'lodash';
 
+const handleResponse = (response) => {
+  // If status is not retrieved, we assume an internal error ocurred, giving the status code 500
+  // Currently 200 is always returned for success responses
+  if (get(response, 'status', 500) !== 200) {
```
luislhl, line 14: Should we compare using `> 299` instead of `!== 200`, to make it future-proof? (in case someone changes the API response codes)

src/api/tokensApi.js
```diff
@@ -8,28 +8,86 @@
...
+   */
+  async getList(searchText, sortBy, order, searchAfter) {
+    const data = {
+      'search_text': searchText,
+      'sort_by': sortBy,
+      'order': order,
+      'search_after': searchAfter.join(','),
+    };
+
+    const response = await requestExplorerServiceV1.get('tokens', { params: data });
```
luislhl, line 46: Can Axios raise some exception while running this, or does it always return something?

src/api/tokensApi.js
```diff
@@ -8,28 +8,86 @@
...
+   */
+  async getBalances(tokenId, sortBy, order, searchAfter) {
+    const data = {
+      'token_id': tokenId,
+      'sort_by': sortBy,
+      'order': order,
+      'search_after': searchAfter.join(',')
+    };
+
+    const response = await requestExplorerServiceV1.get('token_balances', { params: data });
```
luislhl, line 70: Should we set a timeout in the requests? It seems Axios' default is to have no timeout at all.

src/components/token/Tokens.js
```diff
@@ -19,263 +19,293 @@ import ErrorMessageWithIcon from '../error/ErrorMessageWithIcon'
  * Displays custom tokens in a table with pagination buttons and a search bar.
  */
 class Tokens extends React.Component {
+  /**
```
luislhl, line 22: Did you run some kind of formatter in this file? If so, should we make it default for the project, so everyone always use the same rules?

src/components/token/TokenBalances.js
luislhl: > It is not well documented, but does returns a promise
> 
> This code has a race condition without it

Sorry about the other response, I didn't see your comment before.

But I think that's weird. I went to check React's code, and it seems they don't return a Promise indeed: https://github.com/facebook/react/blob/main/packages/react/src/ReactBaseClasses.js#L56

It seems weird to me that this `await` is making any difference. But I believe when you say so, I'm just curious to how exactly this happens, and whether `setState` really returns a Promise somehow.

luislhl: Is it possible that something similar to this is happening in your case?
https://stackoverflow.com/a/47022453/3477266

src/components/token/TokenBalances.js
andreabadesso: Fixed! Thanks

src/components/token/TokenBalances.js
andreabadesso: You are right, refactored the comment on 724795423ba5e17336b98377b740b163543287bd

src/components/token/TokenBalances.js
andreabadesso: You are right, fixed in 2bd55756c14c3e489b1f21238182fb2fe8c0a46c

src/components/token/TokenBalances.js
andreabadesso: You are right, refactored in c67608b9dd1fe17265aecc17822d0fe19e1af2bf

src/components/token/TokenBalances.js
andreabadesso: You are right, that stackoverflows explains why this is working

Refactored in c67608b9dd1fe17265aecc17822d0fe19e1af2bf

src/components/token/TokenBalances.js
andreabadesso: Refactored in c67608b9dd1fe17265aecc17822d0fe19e1af2bf

src/components/token/TokenBalances.js
andreabadesso: Refactored in af10ac6187b2939be1f5942d363677c8f68154bc

src/components/token/TokenBalances.js
andreabadesso: Refactored in af10ac6187b2939be1f5942d363677c8f68154bc

src/api/tokensApi.js
andreabadesso: That's a great idea, refactored in 2009107edd0fd40fa0d0d94a662f13706632d90c

src/api/tokensApi.js
andreabadesso: I've just checked

The axios instance will never throw because we are adding an interceptor and an error handler on `axiosInstance` 

So the `handleResponse` will always receive a response (when the error handler gets an error, it will receive undefined which will default to 500)

src/api/tokensApi.js
andreabadesso: Axios changed it on its latest versions, it used to have a 10s default timeout

We even have a "customAxiosInstance" class that sets this timeout to 30s to be able to load the wallet

I've added a 10s default timeout for token apis in 9034251868aa9ea5a5443a12f58e1596c3dbba87 though

src/components/token/Tokens.js
andreabadesso: I didn't, I manually re-idented it to 2 spaces as the other files have 2 space identation

src/components/token/TokenBalances.js
andreabadesso: I think we can add the selected tokenId, but I suggest doing it in another PR, what do you think?

luislhl (approved):

src/components/token/TokenBalances.js
luislhl: I think in another PR is ok

lucas3003 (dismissed):

src/components/token/TokenBalances.js
andreabadesso: Created an issue for this: https://github.com/HathorNetwork/hathor-explorer/issues/199

andreabadesso:
@luislhl @lucas3003 
I had to fix a typo on the navigation on this commit: 9930d858ebaaf77805636b98b1694fda4d405447

Can you please approve again? Github removed your approvals

Thanks

lucas3003 (approved):

PR #200, opened 2022-06-09
feat: time series screen

lucas3003:
## Acceptance Criteria
Explorer must render the TimeSeries feature inside statistics page with:
- A text stating the last updated data
- The iframe with the Kibana Dashboard

## User Experience

**ATTENTION: THIS  THE DASHBOARD IS NOT SHOWING THE REAL DATA. THERE WILL BE A NEW PR TO REVIEW THE FINAL DASHBOARD**

https://user-images.githubusercontent.com/3287486/173728792-2f6c3e05-6f46-41ce-bf01-c23766f6a94b.mov

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -113,7 +113,6 @@ class Root extends React.Component {
               <NavigationRoute exact path="/transactions" component={TransactionList} />
               <NavigationRoute exact path="/tokens" component={TokenList} />
               <NavigationRoute exact path="/token_balances" component={TokenBalancesList} />
-              <NavigationRoute exact path="/timeseries" component={TimeSeriesDashboard} />
               <NavigationRoute exact path="/blocks" component={BlockList} />
               <NavigationRoute exact path="/dag" component={Dag} />
               <NavigationRoute exact path="/network/:peerId?" component={PeerAdmin} />
--- a/src/App.js
+++ b/src/App.js
@@ -20,7 +20,6 @@ import PushTx from './screens/PushTx';
 import TransactionList from './screens/TransactionList';
 import TokenList from './screens/TokenList';
 import TokenBalancesList from './screens/TokenBalances';
-import TimeSeriesDashboard from './screens/TimeSeriesDashboard';
 import BlockList from './screens/BlockList';
 import TokenDetail from './screens/TokenDetail';
 import Dag from './screens/Dag';
--- a/src/App.js
+++ b/src/App.js
@@ -20,6 +20,7 @@ import PushTx from './screens/PushTx';
 import TransactionList from './screens/TransactionList';
 import TokenList from './screens/TokenList';
 import TokenBalancesList from './screens/TokenBalances';
+import TimeSeriesDashboard from './screens/TimeSeriesDashboard';
 import BlockList from './screens/BlockList';
 import TokenDetail from './screens/TokenDetail';
 import Dag from './screens/Dag';
@@ -112,6 +113,7 @@ class Root extends React.Component {
               <NavigationRoute exact path="/transactions" component={TransactionList} />
               <NavigationRoute exact path="/tokens" component={TokenList} />
               <NavigationRoute exact path="/token_balances" component={TokenBalancesList} />
+              <NavigationRoute exact path="/timeseries" component={TimeSeriesDashboard} />
               <NavigationRoute exact path="/blocks" component={BlockList} />
               <NavigationRoute exact path="/dag" component={Dag} />
               <NavigationRoute exact path="/network/:peerId?" component={PeerAdmin} />
--- a/src/api/blockApi.js
+++ b/src/api/blockApi.js
@@ -1,0 +1,21 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import requestExplorerServiceV1 from './axiosInstance';
+import helpers from '../utils/helpers'
+
+
+const BLOCK_API_DEFAULT_TIMEOUT = 35000; // ms
+
+const blockApi = {
+    async getBestChainHeight() {
+        const response = await requestExplorerServiceV1.get('blocks/best_chain_height', { timeout: BLOCK_API_DEFAULT_TIMEOUT });
+        return helpers.handleExplorerServiceResponse(response);
+    }
+}
+
+export default blockApi;
--- a/src/api/blockApi.js
+++ b/src/api/blockApi.js
@@ -5,13 +5,20 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+import helpers from '../utils/helpers';
+
 import requestExplorerServiceV1 from './axiosInstance';
-import helpers from '../utils/helpers'
 
 
 const BLOCK_API_DEFAULT_TIMEOUT = 35000; // ms
 
 const blockApi = {
+
+    /**
+     * Get the best chain height available on ElasticSearch
+     * 
+     * @returns {Promise}
+     */
     async getBestChainHeight() {
         const response = await requestExplorerServiceV1.get('blocks/best_chain_height', { timeout: BLOCK_API_DEFAULT_TIMEOUT });
         return helpers.handleExplorerServiceResponse(response);
--- a/src/components/Navigation.js
+++ b/src/components/Navigation.js
@@ -79,7 +79,6 @@ class Navigation extends React.Component {
                   <ConditionalNavigation to="/token_balances" label="Token balances" featureToggle={`${UNLEASH_TOKEN_BALANCES_FEATURE_FLAG}.rollout`} />
                 </div>
               </li>
-              <ConditionalNavigation to="/timeseries" label="Time&nbsp;Series" featureToggle={`${UNLEASH_TIME_SERIES_FEATURE_FLAG}.rollout`} />
               <li className="nav-item">
                 <NavLink to="/network" exact className="nav-link" activeClassName="active" activeStyle={{ fontWeight: 'bold' }}>Network</NavLink>
               </li>
--- a/src/components/Navigation.js
+++ b/src/components/Navigation.js
@@ -15,6 +15,7 @@ import ConditionalNavigation from './ConditionalNavigation';
 import {
   UNLEASH_TOKENS_BASE_FEATURE_FLAG,
   UNLEASH_TOKEN_BALANCES_FEATURE_FLAG,
+  UNLEASH_TIME_SERIES_FEATURE_FLAG,
 } from '../constants';
 
 class Navigation extends React.Component {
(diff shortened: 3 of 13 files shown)


src/api/blockApi.js
```diff
@@ -0,0 +1,28 @@
...
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import helpers from '../utils/helpers';
+
+import requestExplorerServiceV1 from './axiosInstance';
+
+
+const BLOCK_API_DEFAULT_TIMEOUT = 35000; // ms
```
pedroferreira1, line 13: Why so high this timeout? What's the API server timeout?

src/components/Navigation.js
```diff
@@ -78,6 +79,7 @@ class Navigation extends React.Component {
                   <ConditionalNavigation to="/token_balances" label="Token balances" featureToggle={`${UNLEASH_TOKEN_BALANCES_FEATURE_FLAG}.rollout`} />
                 </div>
               </li>
+              <ConditionalNavigation to="/timeseries" label="Time&nbsp;Series" featureToggle={`${UNLEASH_TIME_SERIES_FEATURE_FLAG}.rollout`} />
```
pedroferreira1, line 82: Why don't we use a single space instead of `&nbsp;`?

src/utils/helpers.js
```diff
@@ -250,6 +251,21 @@ const helpers = {
    */
   async setStateAsync(instance, state) {
     return new Promise((resolve) => instance.setState(state, resolve));
+  },
+
+  async handleExplorerServiceResponse(response) {
```
pedroferreira1, line 256: Missing docstring

src/utils/date.js
```diff
@@ -11,6 +11,11 @@ const dateFormatter = {
     return `${d.toLocaleDateString('en-US')} ${d.toLocaleTimeString('en-US')}`;
   },
 
+  parseTimestampFromSQLTimestamp(timestamp) {
+    const d = new Date(timestamp);
```
pedroferreira1, line 15: Why don't you use `return parseTimestamp(timestamp / 1000);` ?

You just need to explain it in the docstring but would prevent duplicating the formatting part

src/components/timeseries/TimeSeries.js
```diff
@@ -0,0 +1,38 @@
...
+        this.state = {
+            maintenanceMode: this.props.maintenanceMode
+        };
+    }
+
+    render() {
+        return (
+            (!this.state.maintenanceMode) ?
+                <div>
+                    <ScreenStatusMessage maintenanceMode={this.state.maintenanceMode} />
```
pedroferreira1, line 23: I feel you are not using `maintenanceMode` props in the component, right?

src/components/timeseries/ScreenStatusMessage.js
```diff
@@ -0,0 +1,87 @@
...
+        this.setState({
+            loading: false
+        });
+
+        // Constantly execute the method to get the newest block
+        const screenStatusLoopExecution = setInterval(() => {
+            this.getBestChainHeight();
+        }, 60000); // 60 seconds. This is the interval that ElasticSearch takes to ingest data from blocks
+
+        this.setState({ screenStatusLoopExecution });
```
pedroferreira1, line 35: Maybe we could store this `screenStatusLoopExecution` is `this.screenStatusLoopExecution`, not in state, given that is not used to render anything in the screen.  What do you think?

src/utils/helpers.js
pedroferreira1, line 256: Why is it `async`?

pedroferreira1 (changes requested):
Besides the code review here are some UI/UX suggestions:

1. I think we should have a padding in the screen updated height message. I feel the text is too close to the window and the navigation.
2. I was not thinking about having another navigation link for this to be honest. I feel we should ask Trond about his opinion on this but at first I thought we could use the same Statistics screen but what we have there nowadays we would add a subtitle 'Real time' and you would add your component below in a subtitle 'Historical data', something like that. What do you think?

src/components/timeseries/TimeSeries.js
r4mmer, line 23:
```suggestion
                    <ScreenStatusMessage />
```

`this.state.maintenanceMode` will always be `false` here.
Also, ScreenStatusMessage does not seem to receive any props.

src/components/timeseries/ScreenStatusMessage.js
```diff
@@ -0,0 +1,87 @@
...
+        await this.getBestChainHeight();
+
+        this.setState({
+            loading: false
+        });
+
+        // Constantly execute the method to get the newest block
+        const screenStatusLoopExecution = setInterval(() => {
+            this.getBestChainHeight();
+        }, 60000); // 60 seconds. This is the interval that ElasticSearch takes to ingest data from blocks
```
r4mmer, line 33: Maybe this should be on constants, what do you think?

src/components/timeseries/ScreenStatusMessage.js
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    render() {
+        return (
+            <div>
+                {
+                    (this.state.error) ?
+                        <ErrorMessageWithIcon message='Could not load the last block updated' /> :
+                        (this.state.loading) ?
+                            <Loading /> :
```
r4mmer, line 75: I see this ternary operator with the `Loading` component in many places, maybe we should make a `LoadingArea` component and render either `Loading` or the children based on a prop, but maybe not on this PR since it would change on many places.
What do you think?

src/components/timeseries/ScreenStatusMessage.js
```diff
@@ -0,0 +1,87 @@
...
+
+    componentWillUnmount() {
+        // We need to clear the interval object we created when user leaves the page
+        clearInterval(this.state.screenStatusLoopExecution);
+    }
+
+    /**
+     * Calls the Explorer Service to get the best chain height
+     * 
+     * @returns
```
r4mmer, line 46:
```suggestion
```

src/App.js
```diff
@@ -112,6 +113,7 @@ class Root extends React.Component {
               <NavigationRoute exact path="/transactions" component={TransactionList} />
               <NavigationRoute exact path="/tokens" component={TokenList} />
               <NavigationRoute exact path="/token_balances" component={TokenBalancesList} />
+              <NavigationRoute exact path="/timeseries" component={TimeSeriesDashboard} />
```
r4mmer, line 116: I know that the dashboard is meant to display timeseries charts, but should we show `Time Series` and `/timeseries` to the user? Is there not a more user friendly name for this feature?

r4mmer (changes requested):
I know that ElasticSearch dashboards are easier to configure but I was thinking, should we consider using a grafana instance as a broker for the data?

Grafana has some benefits like:
- The charts would be more customizable
- We would get a better range of graphs to display
- We could mix data from different sources.

And MOST important of all, grafana dashboards can be exported for version control, so we could save the dashboard on this repository and update the instance with the new dashboard during deploy.

src/api/blockApi.js
lucas3003: the Elastic Search timeout is 25 seconds, so the explorer-services times out on 30 seconds. This is why I put 35 seconds. I do agree is too high, but I do not know if this will impact too much the experience, as this is handle async.

src/components/Navigation.js
lucas3003: Will remove this line on the next commit. But, when I was using the single space, the `Time Series` navigation link was being split in two lines.

src/components/timeseries/ScreenStatusMessage.js
```diff
@@ -0,0 +1,87 @@
...
+    /**
+     * Calls the Explorer Service to get the best chain height
+     * 
+     * @returns 
+     */
+    getBestChainHeight = async () => {
+        const blockApiResponse = await blockApi.getBestChainHeight();
+        const blockApiResponseStatus = get(blockApiResponse, 'status', 500);
+
+        if (blockApiResponseStatus > 299) {
```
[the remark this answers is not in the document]
lucas3003: i used that on `blockApi.getBestChainHeight`. But I saw now that I can refactor this method. `handleExplorerServiceResponse` already checks for me if status if >299. I just need to check `blockApiResponse.error` structure. I will refactor in the next commit.

src/components/timeseries/TimeSeries.js
lucas3003: Right. I will remove on the next commit

lucas3003: Right. I was first sending the maintenance mode to this component and let it handle it. But I changed the approach but forgot to remove this part. I will update on the next commit.

src/utils/date.js
lucas3003: We actually receive a date string from ElasticSearch, like `2022-05-09T18:55:47Z`. This is why we could not reuse the code. I will add a comment stating the format we receive.

src/utils/date.js
```diff
@@ -11,6 +11,11 @@ const dateFormatter = {
...
+    return `${d.toLocaleDateString('en-US')} ${d.toLocaleTimeString('en-US')}`;
```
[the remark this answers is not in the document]
lucas3003: For now, to maintain consistency, I think so, because we use the same default on `parseTimestamp` function. But I agree this is not the best UX and we should change this on all experiences inside explorer.

src/utils/helpers.js
lucas3003: Good catch. Do not have to be async

lucas3003: Will put in the next commit.

src/App.js
lucas3003: I will remove it and put on the `Statistics` page

src/components/timeseries/ScreenStatusMessage.js
```diff
@@ -0,0 +1,87 @@
...
+        return (
+            <div>
+                {
+                    (this.state.error) ?
+                        <ErrorMessageWithIcon message='Could not load the last block updated' /> :
+                        (this.state.loading) ?
+                            <Loading /> :
+                            <p>
+                                <strong>
+                                    This screen is updated until block at height {this.state.height} and the last update was on {dateFormatter.parseTimestampFromSQLTimestamp(this.state.timestamp)}
```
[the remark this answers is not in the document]
lucas3003: Done in the last commit

src/components/timeseries/ScreenStatusMessage.js
lucas3003: I agree. I will open a KTLO for this.

src/utils/date.js
pedroferreira1: Ah, now I get it. Maybe add a docstring to explain this and the type of the parameter. I was thinking also about changing the param name, `timestamp` makes me assume it's an integer. However, with the docstring I'm ok to keep the name.

src/components/timeseries/TimeSeries.js
```diff
@@ -0,0 +1,60 @@
...
+    constructor(props) {
+        super(props);
+
+        this.state = {
+            maintenanceMode: this.props.maintenanceMode,
+            featureFlag: this.props.featureFlag,
+        };
+    }
+
+    render() {
```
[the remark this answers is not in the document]
lucas3003: I think this is much better! I will include in the next commit

src/components/timeseries/ScreenStatusMessage.js
```diff
@@ -0,0 +1,79 @@
...
+
+        // Constantly execute the method to get the newest block
+        this.screenStatusLoopExecution = setInterval(() => {
+            this.getBestChainHeight();
+        }, SCREEN_STATUS_LOOP_INTERVAL_IN_SECONDS * 1000);
+    }
+
+    componentWillUnmount() {
+        // We need to clear the interval object we created when user leaves the page
+        clearInterval(this.screenStatusLoopExecution);
```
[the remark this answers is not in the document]
lucas3003: I will include this in the next commit. I think this extra protection is worth to save unnecessary requests to explorer-service

src/utils/date.js
lucas3003: Done

r4mmer (approved):

PR #201, opened 2022-06-13
feat: add token to the query params in the address list screen

pedroferreira1:
### Acceptance criteria

- Address list must have the selected token in the query params, so we can share the link with it.

https://user-images.githubusercontent.com/3298774/173409762-a0ae4f97-cc0a-4d58-9b9f-f3689a1108cc.mov

- We must make it more clear that when there is nothing in the search, then we are showing the address list for HTR

This one is a bit tricky because the best approach would be to have HTR as a search element just like any other token. However, this would require much more changes in the code, then I feel this change is good enough for now with a simple UI change.

image

### DIFF
--- a/src/components/token/TokenAutoCompleteField.js
+++ b/src/components/token/TokenAutoCompleteField.js
@@ -184,7 +184,7 @@ class TokenAutoCompleteField extends React.Component {
         value={this.state.searchText}
         onKeyUp={this.onSearchTextKeyUp}
         onChange={this.onSearchTextChanged}
-        placeholder="Search UID, name or symbol"
+        placeholder="Hathor (HTR) - Type to search for other tokens by UID, name or symbol"
         aria-label="Search"
         ref={ref => {this.autoCompleteInputRef = ref}} />
     )
--- a/src/components/token/TokenAutoCompleteField.js
+++ b/src/components/token/TokenAutoCompleteField.js
@@ -185,8 +185,7 @@ class TokenAutoCompleteField extends React.Component {
         onKeyUp={this.onSearchTextKeyUp}
         onChange={this.onSearchTextChanged}
         placeholder="Hathor (HTR) - Type to search for other tokens by UID, name or symbol"
-        aria-label="Search"
-        ref={ref => {this.autoCompleteInputRef = ref}} />
+        aria-label="Search" />
     )
   }
 
--- a/src/components/token/TokenAutoCompleteField.js
+++ b/src/components/token/TokenAutoCompleteField.js
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import Loading from '../Loading';
 import tokensApi from '../../api/tokensApi';
 import { debounce, get } from 'lodash';
+import { constants as hathorLibConstants } from '@hathor/wallet-lib';
 
 const DEBOUNCE_SEARCH_TIME = 200; // ms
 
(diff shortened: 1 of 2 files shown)


src/components/token/TokenAutoCompleteField.js
```diff
@@ -146,7 +184,7 @@ class TokenAutoCompleteField extends React.Component {
         value={this.state.searchText}
         onKeyUp={this.onSearchTextKeyUp}
         onChange={this.onSearchTextChanged}
-        placeholder="Search UID, name or symbol"
+        placeholder="Hathor (HTR) - Type to search for other tokens by UID, name or symbol"
         aria-label="Search"
         ref={ref => {this.autoCompleteInputRef = ref}} />
```
luislhl, line 189: Do we use this for anything?

I know it's from previous PRs, but maybe we could remove it.

src/components/token/TokenBalances.js
```diff
@@ -83,14 +93,21 @@ class TokenBalances extends React.Component {
     await helpers.setStateAsync(this, {
       sortBy: get(queryParams, 'sortBy', this.state.sortBy),
       order: get(queryParams, 'order', this.state.order),
+      tokenId: get(queryParams, 'token', this.state.tokenId),
       loading: true,
+      queryParamsParsed: true,
```
luislhl, line 98: I think it would be possible to move the line `const queryParams = this.pagination.obtainQueryParams();` to the `constructor`, and set the initial state there already with the right parameters.

This would probably avoid needing to create this `queryParamsParsed` variable.

luislhl (dismissed):

src/components/token/TokenBalances.js
pedroferreira1: Yeah, you are right. Did this refactor here: https://github.com/HathorNetwork/hathor-explorer/pull/201/commits/ba4efc5acf5cb611f38a7c596b2d5b364ca3e26b

src/components/token/TokenAutoCompleteField.js
pedroferreira1: Yeah, it was not used. Removed here https://github.com/HathorNetwork/hathor-explorer/pull/201/commits/1de07de4fef02b6d77ee93c568aed393d7b655df

src/components/token/TokenAutoCompleteField.js
```diff
@@ -56,12 +57,36 @@ class TokenAutoCompleteField extends React.Component {
...
     document.removeEventListener('click', this.handleClick);
   }
 
+  /**
+   * The first component mount will need to find the token item in the elastic search
+   * and then execute the data search, in case a token was already selected in the query
+   *
+   * @param {string} tokenId The token to pre select when first mounting the component
+   */
+  searchAndSelectToken = async (tokenId) => {
```
lucas3003, line 79: `tokenId` is not being used. Either we remove this param or pass it on line 80 instead of `this.props.tokenId`

src/components/token/TokenBalances.js
```diff
@@ -15,6 +15,9 @@ import { withRouter } from "react-router-dom";
 import ErrorMessageWithIcon from '../error/ErrorMessageWithIcon'
 import TokenAutoCompleteField from './TokenAutoCompleteField';
 import helpers from '../../utils/helpers';
+import { constants as hathorLibConstants } from '@hathor/wallet-lib';
+
+
```
lucas3003, line 20: [Nitpick] Too much lines before the class

src/components/token/TokenBalances.js
```diff
@@ -77,20 +90,17 @@ class TokenBalances extends React.Component {
...
-      loading: true,
-    });
-
-    await this.performSearch();
+    if (this.state.tokenId === hathorLibConstants.HATHOR_TOKEN_CONFIG.uid) {
+      // If we had a custom token as queryParam
+      // then we will perform the search after the token
+      // is found in the elastic search
+      // otherwise we just perform the search for HTR to show the default screen
+      await this.performSearch();
```
lucas3003, line 98: I did not get why the performSearch is called only when HTR is the token. Supposing a scenario where a custom token is given as param, where is it going to be searched?

luislhl (dismissed):

src/components/token/TokenBalances.js
pedroferreira1: Thanks. Removed here https://github.com/HathorNetwork/hathor-explorer/pull/201/commits/ee400af32ff7f520d24a543e1842c4612405ac15

src/components/token/TokenAutoCompleteField.js
pedroferreira1: Yeah, I was using this parameter in the beginning, then stopped using and forgot to remove. Thanks.

https://github.com/HathorNetwork/hathor-explorer/pull/201/commits/7c69c887c1f6f8c6e666f5e27b366dce13748cbe

src/components/token/TokenBalances.js
pedroferreira1: Every time you search for a new token in the autocomplete component, a new search is triggered. What I did was to manually select a token in the autocomplete, then the search will be triggered as soon as we get the token element from the elastic search and add to the autocomplete.

I did like this so we don't need to change any structure we have nowadays, the token in the query param just simulate a search and click.

luislhl (approved):

PR #211, opened 2022-07-26
feat: new explorer-service address api

r4mmer:
# Acceptance criteria

- Create client for new address api from explorer service
- Update address details screen to use the new apis


r4mmer:
The new address screen should function exactly the same as the current one.
The main features of this page are:
- Changing tokens on the select reloads the summary and history table.
- Changing the url will select the token and pages accordingly.
- Previous and Next page buttons should work and be disabled if there is no previous or next.
- Searching for a new address will auto select the first token on the list or hathor if it exists on the address' tokens


https://user-images.githubusercontent.com/15909384/181127545-4d285dcf-26d4-41fa-884d-18ad53694665.mp4

OBS: On the video, the summary has some errors because the api behind is still a WIP.

src/api/addressApi.js
```diff
@@ -18,26 +18,33 @@ const addressApi = {
...
   },
 
-  search(address, count, hash, page, token) {
+  getBalance(address, token) {
+    return requestExplorerServiceV1.get(`address/balance`, {params: {address, token}}).then((res) => {
+      if (res && res.data) {
+        return res.data
+      }
+    }).catch((error) => {
+      // something wrong with request
```
lucas3003, line 27: Missing error treatment

src/api/addressApi.js
```diff
@@ -18,26 +18,33 @@ const addressApi = {
...
     /*
      address: address to search
-     count: int -> how many objects we want
-     hash (optional): str -> hash reference for the pagination (works together with 'page' parameter)
-     page (optional): 'previous' or 'next' -> if 'previous', we get the objects before the hash reference
-                                   if 'next', we get the objects after the hash reference
-     token (optional): str -> only fetch txs related to this token uid
+     token: str -> only fetch txs related to this token uid
+     count (optional): int -> how many objects we want
+     skip (optional): str -> skip this many transactions before fetching
```
lucas3003, line 36: This is similart o a `offset`?

r4mmer: Exactly

src/api/addressApi.js
```diff
@@ -18,26 +32,33 @@ const addressApi = {
...
   },
 
-  search(address, count, hash, page, token) {
+  getBalance(address, token) {
+    return requestExplorerServiceV1.get(`address/balance`, {params: {address, token}}).then((res) => {
+      if (res && res.data) {
+        return res.data
+      }
+    }).catch((error) => {
+      // something wrong with request
```
lucas3003, line 41: Missing error treatment

src/components/AddressSummary.js
```diff
@@ -115,15 +115,14 @@ class AddressSummary extends React.Component {
  * address: Address to show summary
  * balance: Object with balance data of each token for the selected address {'uid': {name, symbol, received, spent}}
  * selectedToken: UID of the selected token to show history
- * numberOfTransactions: Number of transactions for the selected token and address
  * tokenSelectChanged: Callback to be executed when user changes token
  */
 AddressSummary.propTypes = {
   address: PropTypes.string.isRequired,
+  tokens: PropTypes.object.isRequired,
```
lucas3003, line 122: add `tokens` to the `propTypes` doc

src/screens/AddressDetail.js
```diff
@@ -13,53 +13,53 @@ import hathorLib from '@hathor/wallet-lib';
...
    * transactions {Array} List of transactions history to show
-   * hasAfter {boolean} If has more elements after the list page
-   * hasBefore {boolean} If has more elements before the list page
    * queryParams {Object} Object with URL parameters data
    * loadingSummary {boolean} If is waiting response of data summary request
    * loadingHistory {boolean} If is waiting response of data history request
+   * loadingTokens {boolean} If is waiting response of tokens request
    * errorMessage {String} message to be shown in case of an error
    * warningRefreshPage {boolean} If should show a warning to refresh the page to see newest data for the address
+   *warnMissingTokens {number} If there are tokens that could not be fetched, this should be the total number of tokens.
```
lucas3003, line 42: missing whitespace

src/screens/AddressDetail.js
```diff
@@ -13,53 +13,53 @@ import hathorLib from '@hathor/wallet-lib';
...
+    loadingSummary: false,
     loadingHistory: false,
+    loadingTokens: true,
     errorMessage: '',
     warningRefreshPage: false,
+    warnMissingTokens: 0,
     selectedTokenMetadata: null,
     metadataLoaded: false,
+    addressTokens: {},
+    txCache: {},
```
lucas3003, line 62: add `txCache` to the documentation above

src/screens/AddressDetail.js
```diff
@@ -178,98 +161,126 @@ class AddressDetail extends React.Component {
...
-      if (response.success) {
-        this.handleFetchedData(response, queryParams);
-      }
-      // fetch metadata for selected token
-      this.getSelectedTokenMetadata(queryParams.token);
-    });
-  }
+  getHistoryData = (page) => {
+    // update query parameters
+    const query = this.pagination.obtainQueryParams()
```
lucas3003, line 166: missing semicolon

r4mmer:
Changed pagination to be more useful with the current pagination scheme.

I chose to show the 2 before and 2 after if they exist and always show the first and last pages.

Screenshot from 2022-08-02 14-48-43
Screenshot from 2022-08-02 14-48-52
Screenshot from 2022-08-02 14-49-03
Screenshot from 2022-08-02 14-49-20

src/api/addressApi.js
r4mmer: Error treatment will be made by the page using the api.

src/api/addressApi.js
r4mmer: Changed to limit/offset to be more intuitive.

src/components/AddressSummary.js
r4mmer: Added

src/screens/AddressDetail.js
r4mmer: Added.

src/screens/AddressDetail.js
r4mmer: Done.

src/screens/AddressDetail.js
r4mmer: Added.

src/api/addressApi.js
r4mmer: Caller will decide how the errors are treated

src/components/AddressDetailExplorer.js
```diff
@@ -0,0 +1,528 @@
...
+  numPages = () => {
+    return Math.ceil(this.state.balance.transactions / TX_COUNT);
+  }
+
+  render() {
+    const renderWarningAlert = () => {
+      if (this.state.warningRefreshPage) {
+        return (
+          <div className="alert alert-warning refresh-alert" role="alert">
+            There is a new transaction for this address. Please <a href="true" onClick={this.refreshPage}>refresh</a> the page to see the newest data.
```
pedroferreira1, line 427: Maybe we could use `refreshPageData` here, what do you think?

src/components/AddressDetailExplorer.js
```diff
@@ -0,0 +1,528 @@
...
+        return null;
+      } else if (this.state.showReloadDataButton || this.state.showReloadTokenButton) {
+          return (
+            <div>
+              <ErrorMessageWithIcon message='The request to get address data has failed. Try to load again, please.' />
+              {renderReloadDataButton()}
+              {renderReloadTokenButton()}
+            </div>
+          );
+      }else {
```
pedroferreira1, line 486: Missing space before `else`

src/components/AddressHistory.js
```diff
@@ -73,27 +45,54 @@ class AddressHistory extends React.Component {
       return <p>This address does not have any transactions yet.</p>
     }
 
-    const getFirstHash = () => {
-      return this.props.transactions[0].tx_id;
+    const paginationLink = (page, query) => {
+        return this.props.pagination.setURLParameters({...query, page: page});
```
pedroferreira1, line 49:
```suggestion
        return this.props.pagination.setURLParameters({ ...query, page });
```

src/components/AddressHistory.js
```diff
@@ -73,27 +45,54 @@ class AddressHistory extends React.Component {
...
+      return Array(1 + end - start).fill().map((_, index) => start + index);
+    };
 
     const loadPagination = () => {
       if (this.props.transactions.length === 0) {
         return null;
       } else {
+        const queryParams = this.props.pagination.obtainQueryParams();
+        const page = +queryParams.page;
+        const lastPage = Math.ceil(this.props.numTransactions / TX_COUNT) - 1;
```
pedroferreira1, line 76: Why `- 1`?

src/components/AddressHistory.js
```diff
@@ -73,27 +45,54 @@ class AddressHistory extends React.Component {
...
-    const getFirstHash = () => {
-      return this.props.transactions[0].tx_id;
+    const paginationLink = (page, query) => {
+        return this.props.pagination.setURLParameters({...query, page: page});
     }
 
-    const getLastHash = () => {
-      return this.props.transactions[this.props.transactions.length-1].tx_id;
-    }
+    const getPages = (page, total) => {
```
pedroferreira1, line 52: We should have more comments explaining in details what this method is doing, I understand that you are selecting which pages will be shown in the pagination, however it's a bit confusing to me. A detailed explanation with examples would be great.

src/components/AddressHistory.js
```diff
@@ -168,9 +167,8 @@ class AddressHistory extends React.Component {
             <td className="d-none d-lg-table-cell pr-3">{hathorLib.helpers.getTxType(tx)}</td>
             <td className="d-none d-lg-table-cell pr-3">{hathorLib.helpers.getShortHash(tx.tx_id)}</td>
             <td className="d-none d-lg-table-cell pr-3">{dateFormatter.parseTimestamp(tx.timestamp)}</td>
-            <td className={tx.is_voided ? 'voided state' : 'state'}>{statusElement}</td>
-            <td>{tx.is_voided && <span className="voided-element">Voided</span>}</td>
```
pedroferreira1, line 172: We don't have the voided information anymore?

src/components/AddressDetailExplorer.js
```diff
@@ -0,0 +1,528 @@
...
+        this.setState({ showReloadTokenButton: true });
+        return;
+      }
+      const txhistory = response || [];
+      this.setState({ transactions: txhistory }, () => {
+
+        const promises = [];
+        for (const tx of txhistory) {
+          if (!this.state.txCache[tx.tx_id]) {
+            promises.push(txApi.getTransaction(tx.tx_id));
```
pedroferreira1, line 191: I would just explain here in details why you need this get. That this is to keep the same column data and in the wallet service you don't have this information.

Explain which column it is and which information you don't have in the wallet service.

src/components/AddressDetailExplorer.js
```diff
@@ -0,0 +1,528 @@
...
+        if (!response) {
+          // An error happened with the API call
+          this.setState({ showReloadDataButton: true });
+          return;
+        }
+
+        let selectedToken = '';
+
+        if (!response) {
+          throw new Error('Error fetching tokens');
```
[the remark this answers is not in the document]
r4mmer: There is a catch at the end that sets the `errorMessage`.

src/components/AddressDetailExplorer.js
```diff
@@ -0,0 +1,528 @@
...
+            // User already have a token selected on the URL
+            this.setState({ selectedToken: queryParams.token }, () => {
+              this.reloadData();
+            });
+          } else {
+            this.reloadData();
+          }
+        });
+      } else {
+        this.setState({ errorMessage: 'Invalid address.' });
```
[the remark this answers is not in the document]
r4mmer: The loading states were not yet set to `true`.

src/components/AddressDetailExplorer.js
```diff
@@ -0,0 +1,528 @@
...
+   * If not valid show error message
+   *
+   * @param {Object} address New searched address to update state
+   */
+  updateAddress = (address) => {
+    this.setState({ queryParams: this.pagination.obtainQueryParams() }, () => {
+      if (hathorLib.transaction.isAddressValid(address)) {
+        this.setState({
+          address,
+          loadingTokens: true,
```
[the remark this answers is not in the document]
r4mmer: While this is true, I believe we should set them on both places.
This would prevent some UI "errors" like showing the table empty and incomplete summary then showing the loading icons.
And the `reloadData` and `reloadTokenData` also need to set the loading states because there are other places which use them (e.g. selected token changed) and having this state management in the method prevents duplicating this code everywhere.

src/components/AddressDetailExplorer.js
```diff
@@ -0,0 +1,528 @@
...
+      loadingSummary: true,
+      loadingHistory: true,
+    }, () => {
+      addressApi.getBalance(this.state.address, token).then(response => {
+        if (!response) {
+          // An error happened with the API call
+          this.setState({ showReloadTokenButton: true });
+          return;
+        }
+        const balance = response || {};
```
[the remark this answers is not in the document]
r4mmer: This was used before I had the if above to prevent api errors.
Our axios instance has an error handler which prevents errors from being thrown, so the `undefined` response is passed in this cases.
Since we already handle this i'll remove it.

src/components/AddressDetailExplorer.js
```diff
@@ -0,0 +1,528 @@
...
+        this.setState({ balance});
+        return balance;
+      }).then(balance => {
+        const query = this.pagination.obtainQueryParams()
+        const page = (query.page && (+query.page >= this.numPages() ? 0 : +query.page)) || 0;
+        return this.getHistoryData(page).then(txhistory => {
+          if (!this.state.metadataLoaded) {
```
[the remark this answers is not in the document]
r4mmer: When I change the selected token for instance, the balance, history and metadata are cleared and we fetch them in order so the metadata will be the last resource fetched.

But if we are reloading from the address, when we fetch the address tokens and select the token to be shown, we already fetch the metadata asynchronously, so when it gets here the metadata may already be loaded.

src/components/AddressHistory.js
r4mmer: An example:

187 tokens with page of length 10 would give us 19 pages (from 0 to 18).
The pages would be: page0 (token1 to token10), ...., page18 (token181 to token187).
But `ceil(187 / 10)` is 19, hence `ceil() - 1` is the last page.

This also works for exact numbers, e.g. 180 tokens would be 18 pages (from 0 to 17) this is why we don't use `floor` since it would give us one more empty page.

src/components/AddressHistory.js
r4mmer: We don't fecth voided transactions with the new API.

src/components/AddressHistory.js
r4mmer: Done.

src/components/AddressHistory.js
r4mmer: Done

src/components/AddressDetailExplorer.js
r4mmer: Done

src/components/AddressDetailExplorer.js
```diff
@@ -0,0 +1,528 @@
...
+          // An error happened with the API call
+          this.setState({ showReloadTokenButton: true });
+          return;
+        }
+        const balance = response || {};
+        this.setState({ balance});
+        return balance;
+      }).then(balance => {
+        const query = this.pagination.obtainQueryParams()
+        const page = (query.page && (+query.page >= this.numPages() ? 0 : +query.page)) || 0;
```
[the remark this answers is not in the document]
r4mmer: Pagination now starts at 1.

src/components/AddressDetailExplorer.js
```diff
@@ -0,0 +1,528 @@
...
+      loadingHistory: true,
+    }, () => {
+      addressApi.getBalance(this.state.address, token).then(response => {
+        if (!response) {
+          // An error happened with the API call
+          this.setState({ showReloadTokenButton: true });
+          return;
+        }
+        const balance = response || {};
+        this.setState({ balance});
```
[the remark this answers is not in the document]
r4mmer: Done

src/components/AddressDetailExplorer.js
```diff
@@ -0,0 +1,528 @@
...
+
+        const tokenDidChange = selectedToken !== this.state.selectedToken || this.state.selectedToken === '';
+
+        this.setState({
+          addressTokens: tokens,
+          loadingTokens: false,
+          selectedToken,
+        }, () => {
+          if (tokenDidChange || !this.state.metadataLoaded) {
+            this.getSelectedTokenMetadata(selectedToken);
```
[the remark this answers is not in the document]
r4mmer: The metadata method now checks for HTR before making the request.

src/api/addressApi.js
```diff
@@ -8,41 +8,56 @@
 import requestExplorerServiceV1 from './axiosInstance';
 
 const addressApi = {
-  getBalance(address) {
-    return requestExplorerServiceV1.get(`node_api/address_balance`, {params: {address}}).then((res) => {
+  getTokens(address, limit, offset) {
+    /*
+     address: address to search tokens on
+     limit (optional): int -> how many objects we want
+     offset (optional): str -> offset this many transactions before fetching
```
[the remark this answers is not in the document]
r4mmer: Fixed

src/screens/AddressDetail.js
```diff
@@ -6,411 +6,31 @@
...
-import helpers from '../utils/helpers';
-import metadataApi from '../api/metadataApi';
-import addressApi from '../api/addressApi';
+import AddressDetailExplorer from '../components/AddressDetailExplorer';
+import AddressDetailLegacy from '../components/AddressDetailLegacy';
+import { useFlag } from '@unleash/proxy-client-react';
+import { UNLEASH_ADDRESS_DETAIL_BASE_FEATURE_FLAG } from '../constants'
+import ErrorMessageWithIcon from '../components/error/ErrorMessageWithIcon'
 
+const AdressDetail = (props) => {
```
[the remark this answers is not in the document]
r4mmer: Fixed

src/constants.js
```diff
@@ -22,6 +22,8 @@ export const WS_URL = tmp_ws_url;
 
 export const TX_COUNT = 10;
 
+export const TOKEN_COUNT = 50;
```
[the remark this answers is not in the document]
r4mmer: Done

src/components/AddressDetailExplorer.js
```diff
@@ -0,0 +1,528 @@
...
+          const hathorUID = hathorLib.constants.HATHOR_TOKEN_CONFIG.uid
+          if (tokens[hathorUID]) {
+            // If HTR is in the token list of this address, it's the default selection
+            selectedToken = hathorUID;
+          } else {
+            // Otherwise we get the first element, if there is one
+            const keys = Object.keys(tokens);
+            if (keys.length === 0) {
+              // In case the length is 0, we have no transactions for this address
+              this.setState({ loadingTokens: false });
```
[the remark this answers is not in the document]
r4mmer: The AddressHistory component shows the message `This address does not have any transactions yet.`.
This message goes in the place where the table would be and is not highlighted like an error message, more like an informational message.

src/components/AddressDetailExplorer.js
```diff
@@ -0,0 +1,528 @@
...
+    });
+  }
+
+  /**
+   * Update URL with new selected token and trigger didUpdate
+   *
+   * @param {String} New token selected
+   */
+  updateTokenURL = (token) => {
+    const newURL = this.pagination.setURLParameters({ token });
```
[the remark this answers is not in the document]
r4mmer: No, the pagination gets the current url and changes only the params we pass, so only the token would be affected.

src/components/AddressDetailExplorer.js
r4mmer: Done

src/components/AddressHistory.js
andreabadesso: Hmmm, was this a design decision? My guess is that we should display voided transactions

It would definitely be possible with the wallet-service backend, we have voided transactions on the `address_tx_history` table

andreabadesso (changes requested):

src/components/AddressDetailExplorer.js
pedroferreira1: Have you tested this? I'm afraid this won't work if you are not in page 1.

You will probably load the same page you are because you will get the page from state, or not?

src/components/AddressHistory.js
pedroferreira1: Yeah, I guess the pagination starting at 0 made the confusion for me.

src/components/AddressDetailExplorer.js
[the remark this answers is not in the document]
r4mmer: Now we show a reload button.

src/components/AddressDetailExplorer.js
r4mmer: Created a method to clear the query parameters then call `refreshPageData` this will make all parameters go to default except the address and selected token, i.e. it will go to the first page of the current token.

src/components/AddressDetailExplorer.js
```diff
@@ -0,0 +1,532 @@
...
+        this.setState({ balance });
+        return balance;
+      }).then(balance => {
+        const query = this.pagination.obtainQueryParams()
+        let page = 1;
+        if (query.page) {
+          if((+query.page > 1) && (+query.page <= this.lastPage())) {
+            page = +query.page;
+          }
+        }
```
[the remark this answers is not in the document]
r4mmer: `+query.page` is a way to convert the page from string to number, it's not an increment.
But I could change to this if you think it's more readable.
```suggestion
        if (query.page) {
          const pageNum = +query.page;
          if((pageNum > 1) && (pageNum <= this.lastPage())) {
            page = pageNum;
          }
        }
```

src/components/AddressHistory.js
pedroferreira1: It was decided with Trond that we won't show voided transactions anymore in this screen

pedroferreira1 (dismissed):

andreabadesso (approved):

pedroferreira1 (approved):

PR #213, opened 2022-08-15
refactor: using transactions count from the token entry instead of the tokenInformation API

andreabadesso:
### Acceptance Criteria
- We should use the new fetch token API from the explorer-service to retrieve the hathor token transactions count
- We should use the `transactions` column from the token entry retrieved from the search to display the total number of transactions for a given token
- If the fetch token API request fails, we should hide the information

### Security Checklist

### DIFF
--- a/src/components/token/TokenBalances.js
+++ b/src/components/token/TokenBalances.js
@@ -236,7 +236,7 @@ class TokenBalances extends React.Component {
 
     this.setState({
       tokensApiError: get(tokenApiRequest, 'error', false),
-      transactionsCount: get(tokenApiRequest, 'data.hits[0].transactions', 0),
+      transactionsCount: get(tokenApiRequest, 'data.hits[0].transactions_count', 0),
     });
   }
 
@@ -256,7 +256,7 @@ class TokenBalances extends React.Component {
 
     await helpers.setStateAsync(this, {
       tokenId: token.id,
-      transactionsCount: token.transactions,
+      transactionsCount: token.transactions_count,
       tokensApiError: false,
     });
 
@@ -285,6 +285,13 @@ class TokenBalances extends React.Component {
    */
   loadingFinished = () => {
     this.setState({ loading: false });
+  }
+
+  onRetryTokensApi = () => {
+    this.setState({
+      tokensApiError: false,
+    });
+    this.fetchHTRTransactionCount();
   }
 
   render() {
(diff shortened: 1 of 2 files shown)


src/components/token/TokenBalances.js
```diff
@@ -226,18 +231,33 @@ class TokenBalances extends React.Component {
     });
   }
 
+  fetchHTRTransactionCount = async () => {
+    const tokenApiRequest = await tokensApi.getToken('00');
```
pedroferreira1, line 235: We should get HTR uid from the lib constant

src/components/token/TokenBalances.js
```diff
@@ -315,7 +335,9 @@ class TokenBalances extends React.Component {
             !this.state.tokenBalanceInformationError && (
               <>
                 <p><b>Total number of addresses:</b> { helpers.renderValue(this.state.addressesCount, true) }</p>
-                <p><b>Total number of transactions:</b> { helpers.renderValue(this.state.transactionsCount, true) }</p>
+                {!this.state.tokensApiError && (
```
pedroferreira1, line 338: Should we write any error message here? So we can let the user know that he can reload the scree to try to get this information.

andreabadesso: I agree, done in 180368e84ebd62af974b4f73f28d639ee0d606dd

src/components/token/TokenBalances.js
andreabadesso: Thanks, done in 7063b34535707b9b049b7aa921bedaec618e73e4

src/components/token/TokenBalances.js
```diff
@@ -311,19 +338,18 @@ class TokenBalances extends React.Component {
...
+
+          {!this.state.tokenBalanceInformationError && (
+            <p><b>Total number of addresses:</b> { helpers.renderValue(this.state.addressesCount, true) }</p>
+          )}
+
+          {!this.state.tokensApiError && (
+            <p><b>Total number of transactions:</b> { helpers.renderValue(this.state.transactionsCount, true) }</p>
+          )}
+
+          {(this.state.tokensApiError || this.state.tokenBalanceInformationError) &&(
```
andreabadesso, line 350: Hides the failed request data and displays a single error message for both

src/components/token/TokenBalances.js
```diff
@@ -261,12 +281,19 @@ class TokenBalances extends React.Component {
...
   /**
    * Turn loading false.
-   * Useful to be used by autocomplete component when the first search doesn't find any token
+  * Useful to be used by autocomplete component when the first search doesn't find any token
    */
   loadingFinished = () => {
     this.setState({ loading: false });
   }
 
+  onRetryTokensApi = () => {
```
pedroferreira1, line 290: Are you using this method?

pedroferreira1 (approved):

src/components/token/TokenBalances.js
andreabadesso: I am not, removed it


Thanks

pedroferreira1 (approved):

pedroferreira1 (approved):

src/components/token/TokenBalances.js
```diff
@@ -311,19 +331,18 @@ class TokenBalances extends React.Component {
...
+          {!this.state.tokenBalanceInformationError && (
+            <p><b>Total number of addresses:</b> { helpers.renderValue(this.state.addressesCount, true) }</p>
+          )}
+
+          {!this.state.tokensApiError && (
+            <p><b>Total number of transactions:</b> { helpers.renderValue(this.state.transactionsCount, true) }</p>
+          )}
+
+          {(this.state.tokensApiError || this.state.tokenBalanceInformationError) &&(
+            <ErrorMessageWithIcon message='Error loading the complete token balance information. Please try again.' />
```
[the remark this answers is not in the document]
andreabadesso: We could, it would be specific for the hathor token though (so is this error message)

I would just suggest opening an issue for this instead of doing it in this PR so we can merge this and fix the bug in production

r4mmer (approved):

Issue #221, opened 2022-08-29
fix: Address detail page when address is empty


r4mmer:
# Summary

When address has no tokens the loading icon does not go away, it should display an error message that the address has no transactions.

PR #222, opened 2022-08-30
feat: show error message for empty addresses

r4mmer:
### Acceptance Criteria
- Show error message when address has no transacions

### Security Checklist

### DIFF
--- a/src/components/AddressDetailExplorer.js
+++ b/src/components/AddressDetailExplorer.js
@@ -225,10 +225,6 @@ class AddressDetailExplorer extends React.Component {
 
         let selectedToken = '';
 
-        if (!response) {
-          throw new Error('Error fetching tokens');
-        }
-
         const tokens = response.tokens || {};
         const total = response.total || 0;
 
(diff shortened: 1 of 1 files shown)


r4mmer:
this closes #221

src/components/AddressDetailExplorer.js
```diff
@@ -225,10 +225,6 @@ class AddressDetailExplorer extends React.Component {
 
         let selectedToken = '';
 
-        if (!response) {
-          throw new Error('Error fetching tokens');
-        }
-
```
r4mmer, line 231: Above we have a conditional for the same, so this will never throw.

src/components/AddressDetailExplorer.js
```diff
@@ -253,7 +249,11 @@ class AddressDetailExplorer extends React.Component {
             const keys = Object.keys(tokens);
             if (keys.length === 0) {
               // In case the length is 0, we have no transactions for this address
-              this.setState({ loadingTokens: false });
+              this.setState({
+                loadingTokens: false,
+                loadingSummary: false,
+                loadingHistory: false,
+              });
```
r4mmer, line 256: This is meant to turn all loading flags off, this way the loading icon will not be shown and the error message can be shown.

andreabadesso (approved):

pedroferreira1 (approved):

PR #230, opened 2022-10-10
refactor: address tx history pagination

andreabadesso:
This PR is part of the https://github.com/HathorNetwork/internal-issues/issues/131 design

### Acceptance Criteria
- We should replace the pages pagination with a `next` and `previous` button
- We should fetch data from the new paginated tx history API

### Security Checklist

### DIFF
--- a/src/components/AddressDetailExplorer.js
+++ b/src/components/AddressDetailExplorer.js
@@ -7,14 +7,14 @@
 
 import React from 'react';
 import AddressSummary from './AddressSummary';
-import AddressHistory from './AddressHistory';
+import AddressHistory from './AddressHistory2';
 import PaginationURL from '../utils/pagination';
 import hathorLib from '@hathor/wallet-lib';
 import ReactLoading from 'react-loading';
 import colors from '../index.scss';
 import WebSocketHandler from '../WebSocketHandler';
 import { TX_COUNT, TOKEN_COUNT } from '../constants';
-import { isEqual } from 'lodash';
+import { isEqual, find } from 'lodash';
 import metadataApi from '../api/metadataApi';
 import addressApi from '../api/addressApi';
 import txApi from '../api/txApi';
@@ -24,7 +24,6 @@ import ErrorMessageWithIcon from '../components/error/ErrorMessageWithIcon'
 class AddressDetailExplorer extends React.Component {
   pagination = new PaginationURL({
     'token': {required: true},
-    'page': {required: false},
   });
 
   addressSummaryRef = React.createRef();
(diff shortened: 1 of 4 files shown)


r4mmer (approved):

src/components/AddressDetailExplorer.js
```diff
@@ -167,26 +175,24 @@ class AddressDetailExplorer extends React.Component {
...
-    const query = this.pagination.obtainQueryParams();
-    if (query.page !== page) {
-      query.page = page;
-      const newURL = this.pagination.setURLParameters(query);
-      this.props.history.push(newURL);
-    }
-
-    return addressApi.getHistory(this.state.address, this.state.selectedToken, TX_COUNT, TX_COUNT * (page-1)).then((response) => {
+  getHistoryData = (lastTx, lastTs) => {
+    console.log('Downloading')
```
luislhl, line 179: Should we remove this?

src/components/AddressDetailExplorer.js
```diff
@@ -167,26 +175,24 @@ class AddressDetailExplorer extends React.Component {
...
+    return addressApi.getHistory(this.state.address, this.state.selectedToken, TX_COUNT, lastTx, lastTs).then((response) => {
       if (!response) {
         // An error happened with the API call
-        this.setState({ showReloadTokenButton: true });
+        this.setState({
+          showReloadTokenButton: true,
```
luislhl, line 184: It seems you used a linter here. Is this linter configured in the project to check all the PRs? Or is it from your IDE?

We should make sure we have a default style and linters that check it on PRs, to avoid everyone using their own linters.

src/components/AddressDetailExplorer.js
```diff
@@ -434,6 +435,56 @@ class AddressDetailExplorer extends React.Component {
...
+    const { timestamp, tx_id } = transactions.at(transactions.length - 1);
+
+    const nextPage = this.state.page + 1;
+
+    const lastTx = tx_id;
+    const lastTs = timestamp;
+
+    // Calculate the next page searchAfter, so it we can click previous and come back to it.
+    const searchAfter = {
+      lastTx,
```
luislhl, line 451: We are not adding the `lastTx` and `lastTs` as parameters in the URL, right?

Didn't the design state that we should update it?

> Also, the URL should change while paginating with the search_after params (the timestamp and tx_id from the last page)

I think it's important to add them in the URL because it makes the URL shareable. What do you think?

luislhl (approved):

src/components/AddressDetailExplorer.js
andreabadesso: Yes we should, removed in 3301af9cb627c140dfd8019387f81792fbad7738

src/components/AddressDetailExplorer.js
andreabadesso: I actually manually changed it, we don't have a default style configured for this project yet

luislhl (approved):

PR #242, opened 2023-07-20
feat(feature-activation): implement features page

glevco:
### Acceptance Criteria

- Implement the new Features page as described in https://github.com/HathorNetwork/rfcs/pull/60

Here are a couple of screenshots of the new page (data is faked):

Screenshot 2023-07-26 at 17 59 50

Screenshot 2023-07-26 at 18 00 10

### Security Checklist
- [x] Make sure you do not include new dependencies in the project unless strictly necessary and do not include dev-dependencies as production ones. More dependencies increase the possibility of one of them being hijacked and affecting us.

PR #244, opened 2023-07-31
feat(feature-activation): implement blocks page

glevco:
Depends on https://github.com/HathorNetwork/hathor-explorer/pull/242

### Acceptance Criteria

- Move `getPrettyState()` to utils file
- Implement the new signal bits table in the Block page as described in the [RFC](https://github.com/HathorNetwork/rfcs/blob/master/projects/feature-activation/0003-explorer-uis.md)

Here's an image of the new implemented block page:

Screenshot 2023-07-31 at 19 49 54

### Security Checklist

### DIFF
--- a/src/api/featureApi.js
+++ b/src/api/featureApi.js
@@ -15,6 +15,13 @@ const featureApi = {
     });
   },
 
+  async getSignalBits(block) {
+    return requestExplorerServiceV1.get(`node_api/feature`, { params: { block } }).then(res => {
+      return res.data.signal_bits
+    });
+  }
+  
+
 };
 
 export default featureApi;
--- a/src/api/featureApi.js
+++ b/src/api/featureApi.js
@@ -9,15 +9,15 @@ import requestExplorerServiceV1 from './axiosInstance';
 
 const featureApi = {
 
-  async getFeatures() {
-    return requestExplorerServiceV1.get(`node_api/feature`).then(res => {
+  async getFeatures(block = null) {
+    return requestExplorerServiceV1.get(`node_api/feature`, { params: { block } }).then(res => {
       return res.data
     });
   },
 
   async getSignalBits(block) {
-    return requestExplorerServiceV1.get(`node_api/feature`, { params: { block } }).then(res => {
-      return res.data.signal_bits
+    return this.getFeatures(block).then(data => {
+      return data.signal_bits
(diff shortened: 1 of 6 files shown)


tuliomir (dismissed):

src/api/featureApi.js
```diff
@@ -15,6 +15,13 @@ const featureApi = {
     });
   },
 
+  async getSignalBits(block) {
+    return requestExplorerServiceV1.get(`node_api/feature`, { params: { block } }).then(res => {
+      return res.data.signal_bits
+    });
```
r4mmer, line 21: This is very similar to the getFeatures above.
Why not have a `getFeatures(block = null)` and have the getSignalBits method call it?

Or maybe use react state to call the api only once and have getSignalBits as a selector of the `signal_bits`.
What do you think?

src/components/feature_activation/SignalBitRow.js
```diff
@@ -0,0 +1,29 @@
...
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+import featureActivation from '../../utils/featureActivation';
+
+
+class SignalBitRow extends React.Component {
+  render() {
+    const { bit, signal, feature, feature_state } = this.props.signalBit
```
r4mmer, line 14: We should be more descriptive, `signalBit` does not seem like a good name for an object with the properties bit, signal, feature, feature_state.
Maybe `featureData` for a `FeatureDataRow` component would be more indicative of the what we will use this component for, what do you think?

src/api/featureApi.js
glevco: Sure! Done in https://github.com/HathorNetwork/hathor-explorer/pull/244/commits/c78ca0f1ff3118c0ac115e302d0f8dc15d8c651e

src/components/feature_activation/SignalBitRow.js
glevco: Sure! Done in https://github.com/HathorNetwork/hathor-explorer/pull/244/commits/c78ca0f1ff3118c0ac115e302d0f8dc15d8c651e

r4mmer (approved):

tuliomir (approved):

r4mmer (approved):

Issue #58, opened 2021-05-20
chore: add JS linter


giovanecosta:
As we are doing a lot of features in explorer and having a lot of comments like "missing semicolon" on PR's, it would be very appreciated if we have a linter to avoid such things.

PR #247, opened 2023-08-03
chore: remove EOL node versions from CI

r4mmer:
### Acceptance Criteria
- Use node18 LST for build and deploy
- Upgrade actions to most recent versions locking the git commit id of the release, not the tag name
- Enable openssl legacy provider to enable build with node18

## Why enable openssl legacy provider

In Node.js v17 a security issue from the SSL provider was fixed, this is a breaking change.

To use node 17 or later we have to either enable the openssl legacy provider flag or upgrade the dependencies to not use the affected libraries.
The upgrade can be done via `npm audit fix --force` but in our case it also required the react-scripts to be upgraded 2 major versions, this new react-scripts uses webpack 5 for build.
Webpack 5 does not automatically include polyfills for node libraries, but we cannot manually add them since react-scripts does not allow custom webpack configs.

The issue can be solved in the wallet-lib by creating a browser compatible build but this requires an upgrade to the wallet-lib version which cannot be done currently since the explorer uses many deprecated apis on the wallet-lib.

The logic solution is to enable the legacy provider, understanding the security issues of openssl and create a project to refactor the explorer for the new wallet-lib version while also making the wallet-lib browser compatible.

## Why remove eslint libs from package.json

We do not use them directly and the version we had fixed required an eslint version with a conflict with the react-scripts required eslint version.
So removing them is the best option.

### Security Checklist


package.json
```diff
@@ -8,8 +8,6 @@
     "bootstrap": "^4.0.0",
     "d3-selection": "^1.3.2",
     "d3-zoom": "^1.7.3",
-    "eslint-config-airbnb": "^16.1.0",
-    "eslint-plugin-react": "^7.5.1",
```
tuliomir, line 12: I agree on removing these since we don't use them directly on this project right now.

But I think it's important to have a linting system available for all our projects. For now, I'll just reinforce the issue that is already open on this matter
- #58

package.json
```diff
@@ -38,7 +36,7 @@
     "watch-css": "npm run build-css && sass --watch src:src",
     "start-js": "react-scripts start",
     "start": "npm-run-all -p watch-css start-js",
-    "build-js": "react-scripts build",
+    "build-js": "react-scripts --openssl-legacy-provider build",
```
tuliomir, line 39: Just double-checking here: no upgrade on dependencies was necessary for this NodeJS "bump", right?

All we needed was to add this `openssl` flag to make `react-scripts` compatible with Node 18.

tuliomir (commented):
Since this PR is dedicated to changing node versions, I'd suggest adding a `.nvmrc` file to make the environment setup easier for the developers. ( [Reference](https://tecadmin.net/lock-application-with-specific-node-version/) )

package.json
r4mmer: react-scripts has eslint as a dependency, so it is already installed, we just need a configuration, which can be done with a package.json key `eslintConfig`.
The best option would be to use the react-scripts default configuration but it may be extended for custom rules.

package.json
r4mmer: Yes, the openssl security issue can be fixed by upgrading the dependencies but it also requires some updates we cannot currently make.
The temporary solution is to enable using the legacy provider until we make the necessary updates.

tuliomir (approved):

glevco (approved):

PR #253, opened 2023-12-07
feat: add support for nano contract details

pedroferreira1:
### Acceptance Criteria
- Add screen to show the details of a nano contract with state and history (TODO the history table still doesn't have pagination but this will be added in the next version).
- Add screen to show details of a blueprint.
- Show nano contract information in the Transaction Detail screen.
- It seems like there's a bug in the `TxData` component when using the `isBlock` method. This PR also fixes it.

https://github.com/HathorNetwork/hathor-explorer/assets/3298774/d268a157-0cbb-4e35-b421-9d0bfc85fb7f

Edit on the screen:

image

### Security Checklist


package.json
```diff
@@ -34,23 +34,17 @@
...
-      "not dead",
-      "not op_mini all"
-    ],
-    "development": [
-      "last 1 chrome version",
-      "last 1 firefox version",
-      "last 1 safari version"
-    ]
-  }
+  "browserslist": [
```
pedroferreira1, line 44: Without this I was getting an error when running locally (with node 18)

src/App.js
```diff
@@ -59,7 +61,11 @@ class Root extends React.Component {
     this.props.apiLoadErrorUpdate({apiLoadError: false});
 
     versionApi.getVersion().then((data) => {
-      hathorLibConfig.setNetwork(data.network.split('-')[0]);
+      let network = data.network;
+      if (data.network.includes('testnet')) {
+        network = 'testnet';
+      }
+      hathorLibConfig.setNetwork(network);
```
pedroferreira1, line 68: The hathor-core is requiring the network to be `nano-testnet-alpha` for testing, so I had to support it with this change.

src/components/tx/TxData.js
```diff
@@ -67,22 +67,41 @@ class TxData extends React.Component {
...
   };
 
   // Array of token uid that was already found to show the symbol
   tokensFound = [];
 
-  componentDidMount = () => {
-    this.calculateTokens();
+  componentDidMount = async () => {
+    await this.handleDataFetch();
   }
```
r4mmer, line 79: Since this is async now, is there any issues with the rendering?
Should we add a loading screen that goes away when this method finishes running?

src/components/tx/TxData.js
```diff
@@ -67,22 +67,41 @@ class TxData extends React.Component {
...
 
-  componentDidUpdate = (prevProps) => {
+  componentDidUpdate = async (prevProps) => {
     if (prevProps.transaction !== this.props.transaction) {
-      this.calculateTokens();
+      await this.handleDataFetch();
     }
```
r4mmer, line 84: If we add the loading screen, we would need to use it here as well

src/screens/nano/NanoContractDetail.js
```diff
@@ -0,0 +1,202 @@
...
+      return (
+        <tr key={name}>
+          <td>{name}</td>
+          <td>{renderAttributeValue(name, data)}</td>
+        </tr>
+      );
+    });
+  }
+
+  // TODO identify that attribute is a token and show as NFT, in case it is.
```
r4mmer, line 185: This TODO is for another PR?

r4mmer (commented):
For the blueprint detail screen I found it somewhat weird to have a table where one of the cells has a table with the arguments.
Maybe we could show the arguments as a method signature, for instance:

| name | arguments | return |
| --- | --- | --- |
| initialize | `(oracle_script: bytes, token_uid: bytes, date_last_bet: int)` | null |

Or something like:

```python
initialize(oracle_script: bytes, token_uid: bytes, date_last_bet: int): None
```

For each method.

tuliomir (commented):
Although it's the nature of the Explorer app to just display data schemas fetched from outside, I felt there could be more references to what data is being displayed here.

This is especially important when displaying the data on the `NanoContractDetail` screen. I think adding some docstrings and explanatory comments that reference the original schemes and designs would be of great help for future code maintenance.

src/App.js
pedroferreira1: I had to do this change also in the desktop wallet. I will create this method in the lib as @tuliomir suggested but I will change here later, otherwise I would need to wait for the lib PR before merging this one here.

When the new version is released I just update here the code.

src/screens/nano/NanoContractDetail.js
pedroferreira1: Yes. Right now we don't have the information when an attribute is a token. I discussed this with Marcelo and we will improve this on hathor-core, so we can better render information on the client.

src/components/tx/TxData.js
pedroferreira1: Makes sense. Added the loading following the UI we use in this screen: https://github.com/HathorNetwork/hathor-explorer/pull/253/commits/39fdb9a17cf350071ae0dfbfa420d710cf118ee9

pedroferreira1:
@tuliomir 

> This is especially important when displaying the data on the NanoContractDetail screen. I think adding some docstrings and explanatory comments that reference the original schemes and designs would be of great help for future code maintenance.

I think having the docs on the APIs is enough. We get the data to show on the screen from the APIs and the user can see it there.

Soon the nano APIs will be in the official API docs, then I can add the links in the code as well.

src/api/nanoApi.js
```diff
@@ -0,0 +1,39 @@
...
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import requestExplorerServiceV1 from './axiosInstance';
+
+const nanoApi = {
+  getState(id, fields, balances, calls) {
```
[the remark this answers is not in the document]
pedroferreira1: Added more details: https://github.com/HathorNetwork/hathor-explorer/pull/253/commits/22b23a6d8192bcfcbb4b828d2e3f0adcc8bb5fb1

tuliomir (approved):

src/components/tx/TxData.js
```diff
@@ -67,20 +67,42 @@ class TxData extends React.Component {
...
+    this.calculateTokens();
+    await this.handleNanoContractFetch();
+  }
+
+  handleNanoContractFetch = async () => {
+    this.setState({ ncLoading: true });
+    if (this.props.transaction.version !== hathorLib.constants.NANO_CONTRACTS_VERSION) {
+      this.setState({ ncLoading: false });
+      return;
     }
```
r4mmer, line 98: There is a possibility of the "nano contract data loading" message flashing for transactions that are not nano contracts, maybe we should use this.

```suggestion
    if (this.props.transaction.version !== hathorLib.constants.NANO_CONTRACTS_VERSION) {
      this.setState({ ncLoading: false });
      return;
    }
    this.setState({ ncLoading: true });
```

r4mmer (approved):

src/components/tx/TxData.js
pedroferreira1: Makes sense: https://github.com/HathorNetwork/hathor-explorer/pull/253/commits/8bdd32b7091d3cc97b22a5564cb6bda00b835565

r4mmer (approved):

PR #262, opened 2024-05-21
fix: nc arguments might be null or an empty array, so we must handle it properly

pedroferreira1:
### Motivation

A beta tester was testing a blueprint that has a method without arguments. A transaction with this method was throwing an error in the frontend because the arguments were `null`. I also improved the UI for an empty list.

### Acceptance Criteria
- We should show a single dash (-) when a nano contract transaction has no arguments in the Transaction Detail screen.

image

### Security Checklist

### DIFF
--- a/src/components/tx/TxData.js
+++ b/src/components/tx/TxData.js
@@ -637,6 +637,10 @@ class TxData extends React.Component {
     }
 
     const renderNCArguments = (args) => {
+      if (!args || !args.length) {
+        return " - ";
+      }
+
       return args.map((arg) => (
         <div key={arg.name}>
           <label>{arg.name}:</label> {renderArgValue(arg)}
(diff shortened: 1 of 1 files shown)


src/components/tx/TxData.js
```diff
@@ -637,6 +637,10 @@ class TxData extends React.Component {
     }
 
     const renderNCArguments = (args) => {
+      if (!args || !args.length) {
```
andreabadesso, line 640:
```suggestion
      if (!Array.isArray(args) || args.length === 0) {
```

andreabadesso (dismissed):

src/components/tx/TxData.js
pedroferreira1: Done https://github.com/HathorNetwork/hathor-explorer/pull/262/commits/c5a2c736554939f855b28b39ccfb989791f1dc89

andreabadesso (approved):

andreabadesso (approved):

PR #274, opened 2024-07-12
ci: deploy to sub-accounts

luislhl:
### Acceptance Criteria
- Use https://github.com/aws-actions/configure-aws-credentials to get temporary AWS Credentials for the job, instead of using permanent credentials stored in Github. The roles that will be assumed by Github Actions for this are being created in https://github.com/HathorNetwork/ops-tools/pull/871
- Add steps to deploy to our alternative testnets every time we deploy to the main one
- Refactor the Makefile so it uses a bash script for the deploy logic

### TODO

- [ ] Remove old user used for deploys

### Security Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,23 @@ ekvilibro_testnet_s3_sync:
 .PHONY: ekvilibro_testnet_deploy
 ekvilibro_testnet_deploy: check_version ekvilibro_testnet_s3_sync clear_cloudfront_cache
 
+.PHONY: ekvilibro_mainnet_build
+ekvilibro_mainnet_build:
+	FULLNODE_HOST=node-side-dag.ekvilibro-mainnet.hathor.network; \
+	export REACT_APP_BASE_URL=https://$$FULLNODE_HOST/v1a/; \
+	export REACT_APP_WS_URL=wss://$$FULLNODE_HOST/v1a/ws/; \
+	export REACT_APP_EXPLORER_SERVICE_BASE_URL=https://explorer-service.ekvilibro-mainnet.hathor.network/; \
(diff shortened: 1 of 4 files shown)


alexruzenhack (dismissed):

Makefile
```diff
@@ -6,6 +6,57 @@ check_version:
...
 
+.PHONY: nano_testnet_build
+nano_testnet_build:
+	FULLNODE_HOST=node1.nano-testnet.hathor.network; \
+	export REACT_APP_BASE_URL=https://$$FULLNODE_HOST/v1a/; \
+	export REACT_APP_WS_URL=wss://$$FULLNODE_HOST/v1a/ws/; \
+	export REACT_APP_EXPLORER_SERVICE_BASE_URL=https://explorer-service.nano-testnet.hathor.network/; \
+	export REACT_APP_TIMESERIES_DASHBOARD_ID=59683ac0-237a-11ef-8f75-578bca86e218; \
+	export REACT_APP_NETWORK=nano-testnet-alpha; \
+	npm run build
```
r4mmer, line 17: suggestion(if-minor): maybe move this to a build script.

The idea is to have all of these parameters pre-configured on the script so we don't repeat this environment configuration on every build command and the command can be a simple `make build site=testnet` and the build script will get the site parameter and use the correct configuration.

Adding others will be simpler since we would just need to add the new site config and build with the `site="new site name"`

What do you think?

luislhl: Done in https://github.com/HathorNetwork/hathor-explorer/pull/274/commits/086c5605c739373d9f323962652e37862b539b47

r4mmer (approved):

alexruzenhack (approved):

r4mmer (approved):

PR #275, opened 2024-07-15
fix v0.19.0 deploy bugs

pedroferreira1:
### Description

1. `TypeError: axios.create is not a function`

It's still unclear why this stopped working in this release but using `import` instead of `require` fixes the bug. https://github.com/HathorNetwork/hathor-explorer/commit/8cabe6fa73f17043dc6d21d25666d6687ef03792

2. hooks can only be called inside of the body of a function component. 

The method `useFlag` can't be used inside of class component. https://github.com/HathorNetwork/hathor-explorer/commit/34f4e6faca16868ce0a8f5b98e452e5d18874c1c

### Acceptance Criteria
- Use `import` instead of `require` for axios.
- Refactor `Navigation` component to a function.

### Security Checklist

### DIFF
--- a/src/components/Navigation.js
+++ b/src/components/Navigation.js
@@ -5,7 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import React from 'react';
+import React, { useRef } from 'react';
 import { NavLink, Link } from 'react-router-dom';
 import logo from '../assets/images/hathor-white-logo.png';
 import HathorAlert from './HathorAlert';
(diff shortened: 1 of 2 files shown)


r4mmer (approved):

src/components/Navigation.js
```diff
@@ -18,111 +18,103 @@ import {
   UNLEASH_TOKEN_BALANCES_FEATURE_FLAG,
 } from '../constants';
 
-class Navigation extends React.Component {
-  constructor(props) {
-    super(props);
+function Navigation(props) {
```
tuliomir, line 21: suggestion(non-blocking): I'd suggest making each of the props explicit, to make it clear what must be passed to this component.
```suggestion
function Navigation({ history }) {
```

In this specific case, making `history` explicit will allow us easily modify it to use the following syntax as soon as we upgrade `react-router-dom` to `v5.x`:
```js
import { useHistory } from "react-router-dom";

function Navigation() {
  const history = useHistory();
  ...
}
```

tuliomir (approved):

src/components/Navigation.js
pedroferreira1: Done https://github.com/HathorNetwork/hathor-explorer/pull/275/commits/05763f6ce5c4f4705d25ddb6b4a3cbb92419adc5

tuliomir (approved):

r4mmer (approved):

PR #287, opened 2024-08-09
feat: show balances of tokens of a nano contract

pedroferreira1:
### Acceptance Criteria
- Show list of balances of a nano contract in the Nano contract detail screen

image
image

### Security Checklist

### DIFF
--- a/src/screens/nano/NanoContractDetail.js
+++ b/src/screens/nano/NanoContractDetail.js
@@ -150,9 +150,9 @@ function NanoContractDetail(props) {
   }
 
   const renderBalances = () => {
-    return Object.entries(ncState.balances).map(([token, data]) => (
-      <tr key={token}>
-        <td>{token === hathorLib.constants.NATIVE_TOKEN_UID ? token : <Link to={`/token_detail/${token}`}>{token}</Link>}</td>
+    return Object.entries(ncState.balances).map(([tokenUid, data]) => (
+      <tr key={tokenUid}>
+        <td>{tokenUid === hathorLib.constants.NATIVE_TOKEN_UID ? tokenUid : <Link to={`/token_detail/${tokenUid}`}>{tokenUid}</Link>}</td>
         <td>{hathorLib.numberUtils.prettyValue(data.value, decimalPlaces)}</td>
       </tr>
     ));
--- a/src/screens/nano/NanoContractDetail.js
+++ b/src/screens/nano/NanoContractDetail.js
@@ -62,7 +62,6 @@ function NanoContractDetail(props) {
         }
 
         const blueprintInformation = await nanoApi.getBlueprintInformation(transactionData.tx.nc_blueprint_id);
-        // TODO get all balances after hathor-core supports it
         const dataState = await nanoApi.getState(ncId, Object.keys(blueprintInformation.attributes), ['__all__'], []);
         if (ignore) {
           // This is to prevent setting a state after the componenet has been already cleaned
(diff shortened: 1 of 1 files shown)


src/screens/nano/NanoContractDetail.js
```diff
@@ -144,6 +144,35 @@ function NanoContractDetail(props) {
...
 
+  const renderBalances = () => {
+    return Object.entries(ncState.balances).map(([token, data]) => {
+      return (
+        <tr key={token}>
+          <td>{data.value}</td>
+          <td>{token === hathorLib.constants.NATIVE_TOKEN_UID ? token : <Link to={`/token_detail/${token}`}>{token}</Link>}</td>
+        </tr>
+      );
+    });
```
r4mmer, line 155:
```suggestion
    return Object.entries(ncState.balances).map(([token, data]) => (
      <tr key={token}>
        <td>{data.value}</td>
        <td>{token === hathorLib.constants.NATIVE_TOKEN_UID ? token : <Link to={`/token_detail/${token}`}>{token}</Link>}</td>
      </tr>
    ));
```

src/screens/nano/NanoContractDetail.js
```diff
@@ -144,6 +144,35 @@ function NanoContractDetail(props) {
...
+      <div className="table-responsive">
+        <table className="table table-striped table-bordered" id="attributes-table">
+          <thead>
+            <tr>
+              <th className="d-lg-table-cell">Amount</th>
+              <th className="d-lg-table-cell">Token</th>
+            </tr>
+          </thead>
+          <tbody>
+            {renderBalances()}
```
r4mmer, line 169: suggestion(if-minor): Instead of using a secondary method, just make `Object.entries(...)` here.

src/screens/nano/NanoContractDetail.js
```diff
@@ -144,6 +144,35 @@ function NanoContractDetail(props) {
...
+  }
+
+  const renderNCBalances = () => {
+    return (
+      <div className="table-responsive">
+        <table className="table table-striped table-bordered" id="attributes-table">
+          <thead>
+            <tr>
+              <th className="d-lg-table-cell">Amount</th>
+              <th className="d-lg-table-cell">Token</th>
```
r4mmer, line 165: suggection(if-minor): Maybe we should invert these.

| Token | Amount |
| --- | --- | 
| 00 | 0.50 | 
| 00112233 | 3.00 |



| Amount | Token |
| --- | --- | 
| 0.50 | 00 | 
| 3.00 | 00112233 | 

The "token first" table seems more intuitive for the reader in my opinion.
What do you think?

src/screens/nano/NanoContractDetail.js
```diff
@@ -144,6 +144,35 @@ function NanoContractDetail(props) {
     });
   }
 
+  const renderBalances = () => {
+    return Object.entries(ncState.balances).map(([token, data]) => {
```
tuliomir, line 148: suggestion(non-blocking): Rename `token` to `tokenUid` or `uid`, to make explicit that it is just a string and not an object.

tuliomir (approved):

src/screens/nano/NanoContractDetail.js
pedroferreira1: The first implementation had the token first but with the size of the Token UID I preferred having the amount first.

I will invert here again and validate my thoughts

src/screens/nano/NanoContractDetail.js
pedroferreira1: Done https://github.com/HathorNetwork/hathor-explorer/pull/287/commits/97222fc0248cf23d79765f40ef45a80154829a75

src/screens/nano/NanoContractDetail.js
pedroferreira1: Done https://github.com/HathorNetwork/hathor-explorer/pull/287/commits/800750ca72d92836d0b086680cf7e129e169a8fc

src/screens/nano/NanoContractDetail.js
pedroferreira1: Done https://github.com/HathorNetwork/hathor-explorer/pull/287/commits/ca4491818941f89c536a414c372ebc777eea98a5

This commit also handles the decimal places to have the pretty value

src/screens/nano/NanoContractDetail.js
pedroferreira1: I usually prefer keeping the map function in a separate method, especially in this file where we have other places that we separate them. Is that ok for you to keep like it is?

r4mmer (approved):

tuliomir (approved):

PR #292, opened 2024-08-15
feat: add list pagination to the nano contract history table

pedroferreira1:
### Acceptance Criteria
- Add pagination to the table of a nano contract history

https://github.com/user-attachments/assets/abdeeafe-ca5a-44e2-a0d4-34309e0f90c9

### Depends on

https://github.com/HathorNetwork/hathor-explorer-service/pull/331
https://github.com/HathorNetwork/nano-hathor-core/pull/95

### Security Checklist

### DIFF
--- a/src/components/nano/NanoContractHistory.js
+++ b/src/components/nano/NanoContractHistory.js
@@ -64,18 +64,22 @@ function NanoContractHistory({ ncId }) {
    */
   const updateListWs = useCallback((tx) => {
     // We only add to the list if it's the first page and it's a new tx from this nano
-    if (!hasBefore) {
-      if (tx.version === hathorLib.constants.NANO_CONTRACTS_VERSION && tx.nc_id === ncId) {
-        let nanoHistory = [...history];
-        const willHaveAfter = (hasAfter || nanoHistory.length === NANO_CONTRACT_TX_HISTORY_COUNT)
-        // This updates the list with the new element at first
-        nanoHistory = helpers.updateListWs(nanoHistory, tx, NANO_CONTRACT_TX_HISTORY_COUNT);
-
-        // Now update the history
-        setHistory(nanoHistory);
-        setHasAfter(willHaveAfter);
-      }
+    if (hasBefore) {
+      return;
     }
+
+    if (tx.version !== hathorLib.constants.NANO_CONTRACTS_VERSION || tx.nc_id !== ncId) {
+      return;
+    }
+
+    let nanoHistory = [...history];
+    const willHaveAfter = (hasAfter || nanoHistory.length === NANO_CONTRACT_TX_HISTORY_COUNT)
+    // This updates the list with the new element at first
+    nanoHistory = helpers.updateListWs(nanoHistory, tx, NANO_CONTRACT_TX_HISTORY_COUNT);
+
+    // Now update the history
+    setHistory(nanoHistory);
+    setHasAfter(willHaveAfter);
   }, [history, hasAfter, hasBefore, ncId]);
 
   /**
--- a/src/components/nano/NanoContractHistory.js
+++ b/src/components/nano/NanoContractHistory.js
@@ -55,9 +55,13 @@ function NanoContractHistory({ ncId }) {
   // hasAfter {boolean} If 'Next' button should be enabled 
   const [hasAfter, setHasAfter] = useState(false);
 
-  // useCallback is important here to update this method with new history state
-  // otherwise it would be fixed the moment the event listener is started in the useEffect
-  // with the history as an empty array
+  /**
+   * useCallback is important here to update this method with new history state
+   * otherwise it would be fixed the moment the event listener is started in the useEffect
+   * with the history as an empty array
+   *
+   * tx {Transaction} Transaction object that arrived from the websocket
+   */
   const updateListWs = useCallback((tx) => {
     // We only add to the list if it's the first page and it's a new tx from this nano
     if (!hasBefore) {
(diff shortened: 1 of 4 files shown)


src/components/nano/NanoContractHistory.js
```diff
@@ -0,0 +1,238 @@
...
+  /**
+   * useCallback is important here to update this method with new history state
+   * otherwise it would be fixed the moment the event listener is started in the useEffect
+   * with the history as an empty array
+   *
+   * tx {Transaction} Transaction object that arrived from the websocket
+   */
+  const updateListWs = useCallback((tx) => {
+    // We only add to the list if it's the first page and it's a new tx from this nano
+    if (!hasBefore) {
```
tuliomir, line 67: suggestion(non-blocking): avoid nested code blocks for simple conditions by early exiting the function
```js
if (hasBefore) {
  return;
}

if (tx.version !== hathorLib.constants.NANO_CONTRACTS_VERSION || tx.nc_id !== ncId) {
  return;
}

// ... actual code and logic here
```

src/components/nano/NanoContractHistory.js
```diff
@@ -0,0 +1,238 @@
...
+
+      if (before) {
+        // We clicked the previous button, so we have next page
+        // and we will have the previous page if has_more === true
+        setHasAfter(true);
+        setHasBefore(data.has_more);
+        if (!data.has_more) {
+          // We are in the first page and clicked the Previous button
+          // so we must clear the query params
+          pagination.clearOptionalQueryParams();
```
tuliomir, line 121: :star: Nice behavior here! Maybe later we could find a way to centralize this on the Pagination util so that other screens also inherit this same pattern without having to explicitly code it on the screen.

src/components/nano/NanoContractHistory.js
```diff
@@ -0,0 +1,238 @@
...
+  }, [location, loadData, pagination]);
+
+  useEffect(() => {
+    // Handle new txs in the network to update the list in real time
+    WebSocketHandler.on('network', handleWebsocket);
+
+    return () => {
+      WebSocketHandler.removeListener('network', handleWebsocket);
+    };
+  });
```
tuliomir, line 160: question: Isn't this missing a `[]` parameter? I would expect the event handler to be initialized only once, and not on every render.
```suggestion
  }, []);
```

src/components/nano/NanoContractHistory.js
tuliomir, line 1: chore: after the recent inclusion of the Eslint configurations on #290, I'd suggest manually running the linter on this file and fixing the issues that should appear.

src/components/nano/NanoContractHistory.js
pedroferreira1: Yes, it's supposed to run only once, thanks https://github.com/HathorNetwork/hathor-explorer/pull/292/commits/526d76871f98e8df03f6f3ab9407c791b5de8792

src/components/nano/NanoContractHistory.js
```diff
@@ -0,0 +1,238 @@
...
+  const [hasBefore, setHasBefore] = useState(false);
+  // hasAfter {boolean} If 'Next' button should be enabled 
+  const [hasAfter, setHasAfter] = useState(false);
+
+  /**
+   * useCallback is important here to update this method with new history state
+   * otherwise it would be fixed the moment the event listener is started in the useEffect
+   * with the history as an empty array
+   *
+   * tx {Transaction} Transaction object that arrived from the websocket
```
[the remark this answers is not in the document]
pedroferreira1: https://github.com/HathorNetwork/hathor-explorer/pull/292/commits/ab25623dc2d2ea2239f12a06330cf8b258aef6bc

src/components/nano/NanoContractHistory.js
```diff
@@ -0,0 +1,238 @@
...
+        setHasAfter(willHaveAfter);
+      }
+    }
+  }, [history, hasAfter, hasBefore, ncId]);
+
+  /**
+   * useCallback is needed here because this method is used as a dependency in the useEffect
+   *
+   * after {string | null} Hash to use for pagination when user clicks to fetch the next page
+   * before {string | null} Hash to use for pagination when user clicks to fetch the previous page
```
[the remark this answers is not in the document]
pedroferreira1: Done, thanks https://github.com/HathorNetwork/hathor-explorer/pull/292/commits/ab25623dc2d2ea2239f12a06330cf8b258aef6bc

src/components/nano/NanoContractHistory.js
pedroferreira1: Much better, thanks https://github.com/HathorNetwork/hathor-explorer/pull/292/commits/993898f82bd11ba930d99bf9458493c078fbffc7

src/components/nano/NanoContractHistory.js
pedroferreira1: There were many more lint errors than I expected, thanks https://github.com/HathorNetwork/hathor-explorer/pull/292/commits/7f04289dda5e4f69ace349f5c0df9ca6574f5ce0

src/components/nano/NanoContractHistory.js
pedroferreira1: Actually after adding the empty array as dependency, the linter complained that I should've added the method `handleWebsocket` as a dependency, because this `useEffect` uses it. With this method as dependency, I had to change it to a `useCallback`, so I had to do one small refactor: https://github.com/HathorNetwork/hathor-explorer/pull/292/commits/960d86bc98537907c965164885b27a7dc125790b

tuliomir (approved):

r4mmer (approved):

PR #294, opened 2024-08-19
style: Manual lint fixes

tuliomir:
### Acceptance Criteria
- Fix all lint warnings and errors on this application
- The only exception is the parsing error found on most screens: that will only be solved by migrating them to functional components

### Security Checklist

### DIFF
--- a/src/api/addressApi.js
+++ b/src/api/addressApi.js
@@ -27,6 +27,7 @@ const addressApi = {
       if (res && res.data) {
         return res.data;
       }
+      return undefined;
     });
   },
 
@@ -37,6 +38,7 @@ const addressApi = {
         if (res && res.data) {
           return res.data;
         }
+        return undefined;
       });
   },
 
@@ -66,6 +68,7 @@ const addressApi = {
       if (res && res.data) {
         return res.data;
       }
+      return undefined;
     });
   },
 };
--- a/src/api/customAxiosInstance.js
+++ b/src/api/customAxiosInstance.js
@@ -17,9 +17,8 @@ import hathorLib from '@hathor/wallet-lib';
  * Create an axios instance to be used when sending requests
  *
  * @param {callback} resolve Callback to be stored and used in case of a retry after a fail
- * @param {number} [_timeout] Timeout in milliseconds for the request
  */
-const createRequestInstance = (resolve, _timeout) => {
+const createRequestInstance = resolve => {
   // Will override lib axios instance increasing the default request timeout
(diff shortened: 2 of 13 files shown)


src/api/addressApi.js
```diff
@@ -27,6 +27,7 @@ const addressApi = {
       if (res && res.data) {
         return res.data;
       }
+      return undefined;
```
r4mmer, line 30:
```suggestion
      return;
```

Would this also work?

src/api/customAxiosInstance.js
```diff
@@ -17,9 +17,9 @@ import hathorLib from '@hathor/wallet-lib';
  * Create an axios instance to be used when sending requests
  *
  * @param {callback} resolve Callback to be stored and used in case of a retry after a fail
- * @param {number} timeout Timeout in milliseconds for the request
+ * @param {number} [_timeout] Timeout in milliseconds for the request
  */
-const createRequestInstance = (resolve, timeout) => {
+const createRequestInstance = (resolve, _timeout) => {
```
r4mmer, line 22: suggestion(if-minor): Should we remove timeout?

If its not used we could just remove this argument, right?

r4mmer (approved):

src/api/addressApi.js
tuliomir: It wouldn't: removing the explicit `undefined` raises an error about an unnecessary `return` on the last line of the function, and removing the `return` breaks the `consistent-return` lint rule that I intended to fix in the first time.

Maybe we can find a better solution at another time, but for now I'd rather not play with code logic.

src/api/customAxiosInstance.js
tuliomir: Good point. Removed on https://github.com/HathorNetwork/hathor-explorer/pull/294/commits/f38cf8ecff200cf9b0fb4764fc167158f13ac1e7

r4mmer (approved):

src/api/metadataApi.js
```diff
@@ -15,8 +15,9 @@ const metadataApi = {
         if (res && id in res.data) {
           return res.data[id];
         }
+        return undefined;
       })
-      .catch(error => {
+      .catch(_error => {
```
andreabadesso, line 20: can't you do:

```suggestion
      .catch(() => {
```

?

andreabadesso (approved):

src/api/metadataApi.js
tuliomir: I'd rather keep the variable, so that we feel more inclined to properly handle these errors in the future

PR #295, opened 2024-08-20
chore: Upgrade Bootstrap from v4 to v5

tuliomir:
### Acceptance Criteria
- The Bootstrap framework should be upgraded from v4 to v5 ( [changes](https://getbootstrap.com/docs/5.0/migration/) )
- There should be a minimum of visual changes

A quick visualization of the visual changes can be seen on the video below.

The tab with the green margins is the one with Bootstrap 5 running on developer localhost, while the one without the border is the existing Bootstrap 4 running on public testnet.

[Screencast from 20-08-2024 20:19:50.webm](https://github.com/user-attachments/assets/8c63203a-3be9-4ff8-bb25-fd936d4f4524)

### Security Checklist

### DIFF
--- a/src/components/Navigation.js
+++ b/src/components/Navigation.js
@@ -76,7 +76,7 @@ function Navigation({ history }) {
           <span className="navbar-toggler-icon" />
         </button>
         <div className="collapse navbar-collapse" id="navbarSupportedContent">
-          <ul className="navbar-nav mr-auto">
+          <ul className="navbar-nav me-auto">
             <li className="nav-item">
               <NavLink
                 to="/"
--- a/src/components/Navigation.js
+++ b/src/components/Navigation.js
@@ -58,7 +58,7 @@ function Navigation({ history }) {
 
   return (
     <div className="main-nav">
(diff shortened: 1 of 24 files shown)


r4mmer (approved):

src/components/Navigation.js
```diff
@@ -165,10 +165,10 @@ function Navigation({ history }) {
               </div>
             </li>
           </ul>
-          <div className="navbar-right d-flex flex-row align-items-center navigation-search">
+          <div className="d-flex flex-row align-items-center ms-auto navigation-search">
```
andreabadesso, line 168: `navbar-right` not needed anymore?

andreabadesso (approved):

src/components/Navigation.js
tuliomir: Not anymore, this behavior is now by default ( [link](https://getbootstrap.com/docs/5.0/components/navbar/#toggler) )

I just added an utility for "margin-right: automatic" so that the search box would stay on the right.

PR #315, opened 2024-09-06
refactor: Root screen to functional

tuliomir:
### Acceptance Criteria
- Refactor Root screen to functional component

### Security Checklist

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -5,9 +5,11 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import React from 'react';
+import React, { useCallback, useEffect } from 'react';
 
 import { Switch, BrowserRouter as Router, Route } from 'react-router-dom';
+import { useDispatch, useSelector } from 'react-redux';
+import { axios as hathorLibAxios, config as hathorLibConfig } from '@hathor/wallet-lib';
 import GDPRConsent from './components/GDPRConsent';
 import Loading from './components/Loading';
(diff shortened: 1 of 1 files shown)


src/App.js
```diff
@@ -35,136 +37,146 @@ import {
...
-    apiLoadErrorUpdate: data => dispatch(apiLoadErrorUpdate(data)),
-    updateServerInfo: data => dispatch(updateServerInfo(data)),
-  };
-};
+const NavigationRoute = ({ internalScreen: InternalScreen }) => (
+  <div>
+    <Navigation />
+    <InternalScreen />
+  </div>
+);
```
r4mmer, line 52: quest(if-minor): Why did we remove `Route` from the `NavigationRoute`?

Since all references of this are inside a `Route` we could just include it here, this makes the router more readable.
What do you think?

r4mmer (approved):

src/App.js
tuliomir: The reason for this is that the structure of Routes in React Router v6 changes, and the guide to migrate from v5.1 recommends this as an intermediate step towards making the upgrade easier.

In short, the idea is that a route should only contain props and data about routing, and not anything related to the business rules of the application. You can read more about it on the [official migration docs](https://reactrouter.com/en/main/upgrading/v5#refactor-custom-routes).

andreabadesso (approved):

PR #317, opened 2024-09-13
refactor: AddressDetailExplorer to fn component

tuliomir:
### Acceptance Criteria
- Refactors the AddressDetailExplorer to functional component

### Motivation
This component used the deprecated [`withRouter` connector](https://v5.reactrouter.com/web/api/withRouter), which will become unavailable in the coming upgrade to React Router v6.

In order to remove its usage, it was necessary to refactor it to use react hooks as a functional component. This proved to be harder than initially foreseen because the code relied heavily upon async state setters, a feature that is also no longer available with `useState()` hooks.

So, instead of setting the state, waiting for it to be rendered and then applying the next set of instructions, the code now sets the state and then a `useEffect()` handles its change. Many data flow reorganizations were made to achieve this new structure.

### Note for the reviewers
Since this component is so complex and had deep changes to its data flow, I'd recommend checking the [before](https://github.com/HathorNetwork/hathor-explorer/blob/b61769d6a6395dbc5c96f5f3c2a462107b693048/src/components/AddressDetailExplorer.js) and [after](https://github.com/HathorNetwork/hathor-explorer/blob/f35fc80fe5aca7b6b4866076081d02567702d8c8/src/components/AddressDetailExplorer.js) versions of the code to understand the changes made.

### Security Checklist

### DIFF
--- a/src/components/AddressDetailExplorer.js
+++ b/src/components/AddressDetailExplorer.js
@@ -5,30 +5,33 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import React from 'react';
+import React, { useCallback, useEffect, useRef, useState } from 'react';
+import hathorLib from '@hathor/wallet-lib';
+import ReactLoading from 'react-loading';
+import { find } from 'lodash';
+import { useHistory, useParams } from 'react-router-dom';
 import AddressSummary from './AddressSummary';
 import AddressHistory from './AddressHistory';
 import PaginationURL from '../utils/pagination';
-import hathorLib from '@hathor/wallet-lib';
-import ReactLoading from 'react-loading';
 import colors from '../index.scss';
 import WebSocketHandler from '../WebSocketHandler';
-import { TX_COUNT, TOKEN_COUNT } from '../constants';
-import { isEqual, find } from 'lodash';
+import { TOKEN_COUNT, TX_COUNT } from '../constants';
 import metadataApi from '../api/metadataApi';
 import addressApi from '../api/addressApi';
 import txApi from '../api/txApi';
-import ErrorMessageWithIcon from '../components/error/ErrorMessageWithIcon';
+import ErrorMessageWithIcon from './error/ErrorMessageWithIcon';
 
-class AddressDetailExplorer extends React.Component {
-  pagination = new PaginationURL({
-    token: { required: true },
-  });
+function AddressDetailExplorer() {
+  const pagination = useRef(
+    new PaginationURL({
+      token: { required: true },
+    })
+  );
 
-  addressSummaryRef = React.createRef();
+  const { address } = useParams();
+  const history = useHistory();
 
   /*
-   * address {String} searched address (from url)
    * selectedToken {String} UID of the selected token when address has many
    * balance {Object} Object with balance of the selected token on this address
    * transactions {Array} List of transactions history to show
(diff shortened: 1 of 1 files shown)


src/components/AddressDetailExplorer.js
```diff
@@ -46,412 +49,362 @@ class AddressDetailExplorer extends React.Component {
...
-      .getHistory(this.state.address, this.state.selectedToken, TX_COUNT, lastTx, lastTs)
-      .then(response => {
-        if (!response) {
-          // An error happened with the API call
-          this.setState({
-            showReloadTokenButton: true,
-          });
-          return;
-        }
+  const getHistoryData = useCallback(
```
r4mmer, line 160: This callback requires `txCache` but it is the only one to call `setTxCache`.
Since this updates the `txCache` it I think we should not use setTxCache and just update the txCache directly.

The idea would be that the txCache is never "set", meaning that the same object will be used between re-renders.
This method adds transactions to the object and the next re-render will have the same object.
What do you think?

src/components/AddressDetailExplorer.js
```diff
@@ -46,412 +49,362 @@ class AddressDetailExplorer extends React.Component {
...
+        await getSelectedTokenMetadata(tokenToReload);
+      } catch (error) {
+        setErrorMessage(error);
       }
-    );
-  };
+      setLoadingSummary(false);
+      return undefined;
+    },
+    [getHistoryData, getSelectedTokenMetadata]
```
r4mmer, line 249: Both of these dependencies can be a simple function, if they are changed to be a funcion maybe this should too.

src/components/AddressDetailExplorer.js
```diff
@@ -46,412 +49,362 @@ class AddressDetailExplorer extends React.Component {
...
-        });
-        return;
-      }
-
-      // update the query params state
-      this.setState({ queryParams, loadingHistory: true });
-    }
-  }
+    },
+  ]);
```
alexruzenhack, line 79: [note] I have a concern about this pattern because it may cause too much re-renderings and it may cause other issues.

alexruzenhack (dismissed):

src/components/AddressDetailExplorer.js
```diff
@@ -46,412 +49,362 @@ class AddressDetailExplorer extends React.Component {
...
+       * Check if the searched address is on the inputs or outputs of the new tx
+       *
+       * @param {Object} tx Transaction data received in the websocket
+       * @param {boolean} checkToken If should also check if token is the same, or just address
+       *
+       * @return {boolean} True if should update the list, false otherwise
+       */
+      function shouldUpdate(tx, checkToken) {
+        const arr = [...tx.outputs, ...tx.inputs];
+        const { token: queryToken } = pagination.current.obtainQueryParams();
```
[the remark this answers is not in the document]
tuliomir: We actually have some logic holes in this component. The way we're calling `shouldUpdate()`, with `checkToken: false`, ensures that `shouldUpdate` will _always_ return `false`. So we never show the alert anyway.

However, to answer your question directly, when we change the token the `onTokenSelectChanged` method is called which basically resets the whole page, so the results would always be correct.

src/components/AddressDetailExplorer.js
```diff
@@ -46,412 +49,362 @@ class AddressDetailExplorer extends React.Component {
...
-              this.reloadData();
+      /**
+       * Check if the searched address is on the inputs or outputs of the new tx
+       *
+       * @param {Object} tx Transaction data received in the websocket
+       * @param {boolean} checkToken If should also check if token is the same, or just address
+       *
+       * @return {boolean} True if should update the list, false otherwise
+       */
+      function shouldUpdate(tx, checkToken) {
```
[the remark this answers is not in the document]
tuliomir: Good point, this refactoring clarification also made a bug stand out.

src/components/AddressDetailExplorer.js
tuliomir: `getHistoryData()` gets called independently of `getSelectedTokenMetadata()` within `onNextPageClicked()` and `onPreviousPageClicked()`.

I think that's why both these functions are separated: there's no need to make an additional metadata call on page change.

src/components/AddressDetailExplorer.js
tuliomir: React batches state changes made within a single synchronous iteration, so there will be only a single re-render for all those changes ( except on some very specific async situations ).

I found [this article](https://medium.com/swlh/react-state-batch-update-b1b61bd28cd2) that explains it way better than I could, with details, example codes and live tests.

src/components/AddressDetailExplorer.js
tuliomir: State immutability is [an important concept for React](https://reacttraining.com/blog/state-in-react-is-immutable). By always using `set` and never mutating the object we can benefit from some features like "time machine" when debugging our application.

However, when studying some other alternative impelmentations for this `getHistoryData` based on your comment, I found a bug that could erase some transactions when under race conditions.

Fixed it on https://github.com/HathorNetwork/hathor-explorer/pull/317/commits/f979f479cbac5b4eeabb1c0410af56e6b9469626

alexruzenhack (approved):

src/components/AddressDetailExplorer.js
r4mmer: My comment had nothing to do with react's state immutability,

The idea was that a change in `txCache` does not seem to change any rendered item.
So a change on the `txCache` would not require a re-render of the screen and the same object could be used between re-renders (for instance, we could use a `useMemo` without dependencies).

**But** after reading more I saw that the only place it is used is to check that a transaction only has authorities (on the history list) and this is updated async so it will require the table to be re-rendered.

src/components/AddressDetailExplorer.js
```diff
@@ -46,412 +49,362 @@ class AddressDetailExplorer extends React.Component {
...
+    [txCache]
+  );
 
-            let selectedToken = '';
+  /**
+   * Fetches the token metadata, if necessary.
+   * Will discard any errors and always set the metadataLoaded to true, even without any data.
+   * @type {(function(*): Promise<void>)|*}
+   */
+  const getSelectedTokenMetadata = useCallback(async tokenToFetch => {
```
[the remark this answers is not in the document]
r4mmer: > To be able to use a function within an effect, we need to define it as a callback

This is not entirely correct, if the funcion definition does not change between re-renders we could just use it on the `useEffect` safely.

But in this case the funcion definition would actually change every re-render so I agree that a `useCallback` is better here.

r4mmer (approved):

PR #327, opened 2024-09-23
fix: Fix QA issues

pedroferreira1:
### Context

After @tuliomir ran the QA, 3 issues were found. They are described [here](https://github.com/HathorNetwork/internal-issues/issues/390#issuecomment-2360848009).

1. The bug in the Transactions list affected all lists, and the cause was that the `componentDidUpdate` was not triggered when the query parameters changed. I decided to refactor the component to a functional component because this would be done in the future anyway, so I fixed the issue and did the refactor.
2. This issue affected all Token Detail screens of NFTs. The fix was simple (@tuliomir had already done it) and was just the correct way of setting the state.
3. This one was caused by a difference in the way we are passing the parameters data to the `Network` component right now. `this.props` had `match: { peerId: undefined }`, so I had to fix the destructuring, so it could correctly get the `peerId` information.

Note: While refactoring the `Transactions` component I found a bug that is affecting the mainnet-production explorer. When the user clicks the 'Previous' button in the list, the API returns the data in the reverse order, so we must reverse the data when this happens. In the middle of the refactor, I also fixed it.

### Acceptance Criteria
- Refactor `Transactions` component to a functional component fixing the bug when the update was not being triggered when the query params changed.
- Fix order of data when user clicked the 'Previous' button in lists that used the `Transactions` component.
- Fix Token Detail screen when the token is an NFT correctly setting the state of metadata.
- Fix Network screen correctly getting the `peerId` query param.

### Security Checklist

### DIFF
--- a/src/components/tx/Transactions.js
+++ b/src/components/tx/Transactions.js
@@ -5,16 +5,16 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import React from 'react';
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
 import ReactLoading from 'react-loading';
-import { Link } from 'react-router-dom';
-import { isEqual } from 'lodash';
+import { Link, useLocation } from 'react-router-dom';
 import { TX_COUNT } from '../../constants';
 import TxRow from './TxRow';
 import helpers from '../../utils/helpers';
 import WebSocketHandler from '../../WebSocketHandler';
 import colors from '../../index.scss';
 import PaginationURL from '../../utils/pagination';
+import { reverse } from 'lodash';
 
 /**
  * Displays transactions history in a table with pagination buttons. As the user navigates through the history,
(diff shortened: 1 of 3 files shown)


src/screens/TokenDetail.js
```diff
@@ -79,11 +79,9 @@ function TokenDetail() {
     }
     setToken(oldToken => {
       return {
-        token: {
-          ...oldToken,
-          uid: id,
-          meta: data,
-        },
+        ...oldToken,
+        uid: id,
```
r4mmer, line 83: question(if-minor): Should we stop resetting the id?

src/components/tx/Transactions.js
```diff
@@ -32,233 +32,273 @@ import PaginationURL from '../../utils/pagination';
...
-      lastTimestamp,
-      hasAfter,
-      hasBefore,
-      queryParams,
-    });
-  };
+  useEffect(() => {
+    // Handle load history depending on the query params in the URL
+    getData(pagination.obtainQueryParams());
+  }, [location, getData, pagination]);
```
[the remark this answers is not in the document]
pedroferreira1: The trigger happens because of the `location` but the other is to make the compiler stops complaining.

The same as in the previous comment.

src/components/tx/Transactions.js
```diff
@@ -32,233 +32,273 @@ import PaginationURL from '../../utils/pagination';
  *   ts = "1579637190"
  *   page = "next"
  */
-class Transactions extends React.Component {
-  constructor(props) {
-    super(props);
+function Transactions({ shouldUpdateList, updateData, title }) {
+  // We must use memo here because we were creating a new pagination
+  // object in every new render, so the useEffect was being called forever
+  const pagination = useMemo(
```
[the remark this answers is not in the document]
pedroferreira1: Make sense, done https://github.com/HathorNetwork/hathor-explorer/pull/327/commits/05f8424a42bdf6248fb2977c46e6584727c3c693

src/components/tx/Transactions.js
```diff
@@ -32,233 +32,273 @@ import PaginationURL from '../../utils/pagination';
  *   ts = "1579637190"
  *   page = "next"
  */
-class Transactions extends React.Component {
-  constructor(props) {
-    super(props);
+function Transactions({ shouldUpdateList, updateData, title }) {
+  // We must use memo here because we were creating a new pagination
+  // object in every new render, so the useEffect was being called forever
```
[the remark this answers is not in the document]
pedroferreira1: I agree that it should be different from what it is nowadays. Maybe we could have an event thrown when parameters change. I will create an issue to discuss that, the idea of this PR is just to refactor the component to fix the bugs found.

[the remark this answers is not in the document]
pedroferreira1: https://github.com/HathorNetwork/hathor-explorer/issues/328

src/screens/TokenDetail.js
pedroferreira1: I believe we should but this screen should be redesigned. It is like this because there are two async requests and both of them set the uid. If this is the first response, then it must set the uid correctly.

I think the screen should be re-implemented in a way that it shows the screen after both requests finish, then we have the completed information and we don't need to set the id twice.

I prefer keeping like this and only fix the bug. What do you think?

r4mmer: Makes sense

src/components/tx/Transactions.js
```diff
@@ -32,233 +32,263 @@ import PaginationURL from '../../utils/pagination';
...
+    getData(pagination.obtainQueryParams());
+  }, [location, getData, pagination]);
+
+  useEffect(() => {
+    // Handle new txs in the network to update the list in real time
+    WebSocketHandler.on('network', handleWebsocket);
+
+    return () => {
+      WebSocketHandler.removeListener('network', handleWebsocket);
     };
```
alexruzenhack, line 213: [question] Is this unmounting really working? I'm asking this because React has probably a bug when dealing with unmounting when more then 1 `useEffect` is declared in the component. If it is working, all good, otherwise I would solve this issue by declaring this `useEffect` first.

pedroferreira1: I tested and it's working fine when the screen is unmounted. What's this bug that React has with more than one `useEffect`?

alexruzenhack: Maybe this is a thing of React Native. When we use two `useEffect` with `unmounting` callbacks only the the first `unmounting` callback from the first `useEffect` is called, never the second. This case is little different because the first `useEffect` didn't register an `unmounting`, maybe you can make this test latter. However, for this PR it is enough. I'm approving.

alexruzenhack (approved):

PR #334, opened 2024-10-16
feat: revamp home ui

julia-eileen:
### Acceptance Criteria
- Configure the project to allow the old UI codebase to coexist with the new one until the revamp is completed.
- Implement a feature flag system using the useNewUiEnabled hook.
- Modify the Sass build script to generate output for both the old and new UI.
- Set up the project to allow a theming system (dark and light mode).
- Update navigation, status components, and the homepage according to the new UI.

### Security Checklist

https://github.com/user-attachments/assets/11514966-ad9a-42cc-942e-708a831b8db1

https://github.com/user-attachments/assets/6673bec3-dd17-4324-b47b-bf2b18e7f4b1


julia-eileen:
To quickly test the new UI without modifying the feature flag response, simply update the `hooks/useNewUiEnabled.js` file with the following content:

```
/**
 * Copyright (c) Hathor Labs and its affiliates.
 *
 * This source code is licensed under the MIT license found in the
 * LICENSE file in the root directory of this source tree.
 */

import { useFlag } from '@unleash/proxy-client-react';
import { UNLEASH_NEW_UI_FEATURE_FLAG } from '../constants';

export const useNewUiEnabled = () => {
  const newUiEnabled = useFlag(UNLEASH_NEW_UI_FEATURE_FLAG) || true;

  return newUiEnabled;
};
```

src/screens/DashboardTx.js
```diff
@@ -99,26 +108,110 @@ function DashboardTx() {
...
-    </div>
-  );
+    );
+  };
+
+  const renderNewTables = (content, text, path) => {
+    return (
+      <div style={{ textAlign: 'center' }}>
+        <div>
+          <table className="table-stylezed">
```
tuliomir, line 138: fix: minor typo. Maybe there are other similar classes with the same fix.
```suggestion
          <table className="table-stylized">
```

src/components/Switch.js
```diff
@@ -0,0 +1,34 @@
...
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React, { useState } from 'react';
+import { useDispatch, useSelector } from 'react-redux';
+import { useMount } from '../hooks/useMount';
+import { toggleTheme } from '../actions';
+
+export const SwitchButton = () => {
+  const [value, setvalue] = useState(false);
```
tuliomir, line 14: chore: minor typo on camelCase convention
```suggestion
  const [value, setValue] = useState(false);
```

src/components/Switch.js
```diff
@@ -0,0 +1,34 @@
...
+    }
+  });
+
+  const changeValue = () => {
+    setvalue(!value);
+    dispatch(toggleTheme());
+  };
+
+  return (
+    <div className={`switch ${value ? 'active' : ''}`} onClick={() => changeValue(!value)}>
```
tuliomir, line 30: chore: We are sending a parameter to `changeValue()` that is unused.

src/actions/index.js
```diff
@@ -15,3 +17,13 @@ export const isVersionAllowedUpdate = data => ({
 export const apiLoadErrorUpdate = data => ({ type: 'api_load_error_update', payload: data });
 
 export const updateServerInfo = data => ({ type: 'update_server_info', payload: data });
+
+export const toggleTheme = () => {
```
tuliomir, line 21: suggestion: This action function has side effects even before the actual payload is dispatched.

While this makes sense in this context, it may create confusion and bugs in future maintenances. And it also intrigues me that the action decision is based on the `html` theme, and not on Redux store data.

My initial suggestion would be creating an utils function that implement the side effects while also dispatching the `toggle_theme` action. This would make it clear to the function caller what is the expected result from the get go and avoid context interference.

Do you have other ideas on how to better structure this change?

src/components/GDPRConsent.js
```diff
@@ -10,33 +10,80 @@ import React, { useEffect } from 'react';
 import CookieConsent from 'react-cookie-consent';
 import TagManager from 'react-gtm-module';
 import { GTM_ID } from '../constants';
+import { useNewUiEnabled } from '../hooks';
 
 const GDPRConsent = () => {
+  const newUiEnabled = useNewUiEnabled();
+
+  const style = {
```
tuliomir, line 18: suggestion(non-blocking): Minor suggestion to make the variable clearer
```suggestion
  const newUiStyle = {
```

src/components/Navigation.js
```diff
@@ -57,26 +81,44 @@ function Navigation() {
...
+            <div className="d-flex flex-column align-items-center">
+              <Link className="navbar-brand" to="/" href="/">
+                <NewLogo
+                  className={`newLogo ${theme === 'dark' ? 'dark-theme-logo' : 'light-theme-logo'}`}
+                />
+              </Link>
+            </div>
+            {!showSearchInput ? (
+              <button
+                className="navbar-toggler {}"
```
tuliomir, line 98: question: Is the className `{}` necessary?

tuliomir (commented):
Thanks for contributing to our project! Here are some questions and suggestions on an initial partial review.

src/utils/getTheme.js
```diff
@@ -0,0 +1,27 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+const getTheme = () => {
```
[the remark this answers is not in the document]
julia-eileen: I have refactored it to the following functions:

getStoredTheme: Retrieves the user's stored theme preference from localStorage.
getSystemTheme: Determines the system's color scheme using window.matchMedia.
applyTheme: Applies the given theme to the HTML element's dataset and saves it to localStorage.
initializeTheme: Combines the logic of checking stored preferences, determining the system's default, and applying the theme accordingly.

Each function is now focused on a single responsibility, with initializeTheme coordinating the overall flow.

src/hooks/useMount.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { useEffect } from 'react';
+
+export const useMount = cb => {
```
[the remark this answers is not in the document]
julia-eileen: Okis! 

We've created the useMount hook because it makes it very clear that the function will only run on the component's mount. It's easier to understand at a glance compared to manually writing useEffect with an empty dependency array. Futhermore one of the key benefits of the useMount hook is that it avoids the ESLint warning about missing dependencies in the useEffect hook. While the warning can be disabled using comments, that approach adds clutter to the code. Using useMount abstracts this detail away and avoids the need to justify the absence of a dependencies array repeatedly.

But removed it.

src/actions/index.js
julia-eileen: The utility function would require access to the dispatch function, which could complicate the code further. The action currently uses the previous Redux state to toggle the active theme. We rely on the theme string stored in the Redux state for conditional styling in components. However, it's also necessary to set the `data-theme` attribute on the HTML element so that the CSS variables work correctly. We've made some refactorings—please review and see if they address the concerns.

src/components/Navigation.js
julia-eileen: removed.

src/components/Switch.js
julia-eileen: Fixed.

src/screens/DashboardTx.js
julia-eileen: Fixed as suggested.

src/hooks/useNewUiLoad.js
```diff
@@ -0,0 +1,31 @@
...
+import { useNewUiEnabled } from './useNewUiEnabled';
+
+export const useNewUiLoad = () => {
+  const [loading, setLoading] = useState(true);
+  const newUiEnabled = useNewUiEnabled();
+
+  useEffect(() => {
+    if (!newUiEnabled) {
+      setLoading(false);
+    } else {
```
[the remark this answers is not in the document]
julia-eileen: Fixed as suggested.

src/hooks/useIsMobile.js
```diff
@@ -0,0 +1,23 @@
...
+ */
+
+import { useState, useEffect } from 'react';
+
+export const useIsMobile = () => {
+  const [isMobile, setIsMobile] = useState(window.innerWidth < 768);
+
+  useEffect(() => {
+    const handleResize = () => {
+      setIsMobile(window.innerWidth < 850);
```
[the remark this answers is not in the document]
julia-eileen: Fixed as suggested.

src/hooks/useMount.js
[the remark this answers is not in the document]
tuliomir: In that `Switch.js` case, maybe an appropriate solution would be to use the default value declaration and remove the `useEffect` altogether:

```jsx
  const theme = useSelector(state => state.theme);
  const [value, setValue] = useState(theme === 'dark');
```

We try to follow the recommendations from React conventions and this `react-hooks/exhaustive-deps` is reinforced by React devs themselves in many places ( [example](https://github.com/facebook/create-react-app/issues/6880#issuecomment-485912528) ).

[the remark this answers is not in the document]
julia-eileen: Each storage location has its own responsibility. LocalStorage is intended for persisting the user’s choice, but it's not meant to be used across the application. For application-wide state management, we rely on the existing Redux store, which ensures a consistent data flow throughout the project.

Regarding the data-attribute in HTML, it's specifically used for storing CSS theme variables because the project doesn't utilize a CSS-in-JS solution.

In the switch component, local state was being used to manage a true/false value, but it can be refactored to directly use the theme value in the conditional logic, simplifying the component.

As for the exhaustive-deps, we’re not suggesting removing it. The intent was to apply a common pattern where the effect runs only on mount, without re-triggering unnecessarily. This approach helps when you only need to set values once during the component lifecycle.

Refactored the component to rely only in the theme saved in the store.

tuliomir (commented):
Not related to any file in specific: When running the application in localhost using the new UI, I'm receiving the following warning on browser console:

```
index.js:26 Warning: validateDOMNesting(...): <li> cannot appear as a descendant of <li>.
    at li
    at ConditionalNavigation (http://localhost:3000/static/js/main.chunk.js:4765:5)
```

This may be related to some specific usage on the `Navigation.js` file, but I'm unsure of its source as of now.

julia-eileen:
> Not related to any file in specific: When running the application in localhost using the new UI, I'm receiving the following warning on browser console:
> 
> ```
> index.js:26 Warning: validateDOMNesting(...): <li> cannot appear as a descendant of <li>.
>     at li
>     at ConditionalNavigation (http://localhost:3000/static/js/main.chunk.js:4765:5)
> ```
> 
> This may be related to some specific usage on the `Navigation.js` file, but I'm unsure of its source as of now.

It's because of the dropdown, can we fix it in the next PR? it would take some time to refactor this.

tuliomir (dismissed):

src/components/Navigation.js
```diff
@@ -169,22 +213,196 @@ function Navigation() {
...
+            data-bs-toggle="collapse"
+            data-bs-target="#navbarSupportedContent"
+            aria-controls="navbarSupportedContent"
+            aria-expanded="false"
+            aria-label="Toggle navigation"
+          >
+            <span className="navbar-toggler-icon" />
+          </button>
+          <div className="collapse navbar-collapse" id="navbarSupportedContent">
+            <ul className="navbar-nav me-auto">
```
r4mmer, line 295: Correct me if i'm wrong, but it seems there is a lot of duplicated items in the `renderUi`, `renderNewUi` and `SideBar`.
Maybe we should create a component for these duplicated items.

For instance, this list (`ul`) seems like it would have no problem being extracted to another component.
What do you think?

src/components/Navigation.js
```diff
@@ -169,22 +213,196 @@ function Navigation() {
           <div className="d-flex flex-row align-items-center ms-auto navigation-search">
             <div className="d-flex flex-row align-items-center">
               <input
-                className="form-control me-2"
+                className="form-control me-2 bg-dark text-light navigation-search-input"
                 type="search"
-                placeholder="Search tx or address"
+                placeholder={`Search here`}
```
r4mmer, line 218: issue: The UI is not indicative of what can be searched.

We should keep the "Search tx or address" so users know that he can also search for an address.
Or maybe we can keep the "Search here" and when the user focus on the input it shows a message explaining that he can search both txs and addresses.

src/components/Navigation.js
```diff
@@ -169,22 +213,196 @@ function Navigation() {
...
-            <Version />
           </div>
         </div>
+        <div className="mainnet-container hide-tabs">
+          <img
+            src={theme === 'dark' ? sun : moon}
+            alt="themeColorButton"
+            className="theme-color-btn"
+            onClick={() => dispatch(toggleTheme())}
+          />
```
r4mmer, line 232: suggestion: Move this to the theme toggle component.

Maybe the theme toggle component should receive a bool to render as a switch or an image button.
This way we don't have to inline the theme toggle.

src/components/Navigation.js
```diff
@@ -169,22 +213,196 @@ function Navigation() {
...
+              onBlur={() => setShowSearchInput(false)}
+              autoFocus
+            />
+          ) : (
+            <SearchIcon
+              fill={theme === 'dark' ? 'white' : 'black'}
+              onClick={toggleSearchInput}
+              className="mobile-search-icon"
+            />
+          )}
```
r4mmer, line 262: Suggestion: Extract this into another component.

Just an idea, since `showSearchInput` can be moved to the component and the rest seems to work.
Also, this Navigation component would be more straightforward and easier to read.

src/components/Switch.js
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+import { useDispatch, useSelector } from 'react-redux';
+import { toggleTheme } from '../actions';
+
+export const SwitchButton = () => {
```
r4mmer, line 12: issue: We should have the same

src/components/tx/TxRow.js
```diff
@@ -9,15 +9,60 @@ import React from 'react';
...
+        {idStart}
+        <div className="ellipsis">
+          <div className="ellipsi"></div>
+          <div className="ellipsi"></div>
+          <div className="ellipsi"></div>
+        </div>
+        {idEnd}
+      </div>
+    );
+  };
```
r4mmer, line 56: suggestion: Maybe we should move this to the utils folder so we can use this timestamp parser in other places as well.

julia-eileen: We moved it, in the next PR it will be in the utils/date.js

src/components/Navigation.js
julia-eileen: We changed it accordingly with the figma prototype. 

Screenshot_20241021_171735_Firefox

Should we change it back?

src/App.js
```diff
@@ -65,6 +76,18 @@ function Root() {
...
+    return newUiEnabled ? (
+      <ErrorMessage />
+    ) : (
+      <div className="content-wrapper">
+        <h3 className="text-danger">
+          Error loading the explorer. Please reload the page to try again.
+        </h3>
+      </div>
+    );
+  };
```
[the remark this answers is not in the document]
julia-eileen: Fixed as sugested.

src/components/Navigation.js
```diff
@@ -5,27 +5,51 @@
...
+import React, { useRef, useState } from 'react';
 import { NavLink, Link, useHistory } from 'react-router-dom';
 import hathorLib from '@hathor/wallet-lib';
 import { useFlag } from '@unleash/proxy-client-react';
+import { useDispatch, useSelector } from 'react-redux';
+import { useNewUiEnabled } from '../hooks';
 import logo from '../assets/images/hathor-white-logo.png';
+import moon from '../assets/images/moon.svg';
+import sun from '../assets/images/sun-dark.svg';
+import { ReactComponent as NewLogo } from '../assets/images/new-logo.svg';
```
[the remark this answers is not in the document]
julia-eileen: When importing an SVG as a React component, it's easier to modify properties like fill and stroke to match the theme's styling. However, when using the SVG as a background, we often need to import both versions, since we can't dynamically adjust the SVG properties in that context.

src/components/Navigation.js
```diff
@@ -57,26 +81,44 @@ function Navigation() {
...
+              ''
+            )}
+          </div>
+          <button className="hide-mainnet-mobile">
+            <GlobeMainnet
+              className={`${
+                theme === 'dark' ? 'dark-theme-logo' : 'light-theme-logo'
+              } theme-mainnet-btn`}
+            />
+            <span className="nav-title">Hathor Mainnet</span>
```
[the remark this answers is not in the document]
julia-eileen: How should I use the network.getNetwork? It gives me an object with alias and name, the alias I receive just `test` and the name `htr-testnet`, should I use it like that:

```
const navTitle = `Hathor ${network.getNetwork().alias}`; // Hathor Test
```

or

```
const navTitle = `${network.getNetwork().name}`; // Htr-testnet
```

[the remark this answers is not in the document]
julia-eileen: **1st option:**
image

**2nd option:**
image

src/components/Navigation.js
julia-eileen: I agree, but since we focused on customizing the existing UI for the proposed revamp, aiming to minimize changes (even in terms of componentization), could we consider postponing this refactor for now?

src/components/Navigation.js
julia-eileen: I refactored the JSX to make the roles of the elements clearer. Could you please review it?

src/components/Switch.js
julia-eileen: Fixed as sugested.

src/components/Navigation.js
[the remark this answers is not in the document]
julia-eileen: I've used the `name` property, but if this isn't the best approach, let me know and I'll update it.

src/components/Navigation.js
```diff
@@ -41,6 +41,7 @@ function Navigation() {
   const [showSearchInput, setShowSearchInput] = useState(false);
   const [showSidebar, setShowSidebar] = useState(false);
   const showTokensTab = isTokensBalanceEnabled || isTokensBaseEnabled;
+  const network = hathorLib.config.getNetwork();
```
r4mmer, line 44: I'm sorry I wasn't more clear on the last review.
I forgot the explorer is built with the `REACT_APP_NETWORK` constant that holds the network.

We can just import `REACT_APP_NETWORK` from '../constants', it will have `mainnet` or `testnet` we want to use. ([reference](https://github.com/julia-eileen/hathor-explorer/blob/feat/revamp-home-ui/src/constants.js#L12)).

src/components/Navigation.js
r4mmer: Sure!

src/components/Navigation.js
r4mmer: > We changed it accordingly with the figma prototype.

You're right, I think we can merge like this and we can talk with Ana later since this change is simple to rollback.

src/components/Navigation.js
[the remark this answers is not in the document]
r4mmer: I checked the app constanst and found a better way to do this, explanation in this [comment](https://github.com/HathorNetwork/hathor-explorer/pull/334#discussion_r1811015329)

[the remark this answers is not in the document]
julia-eileen: Thanks, fixed it using the constant value now.

r4mmer (approved):

tuliomir (approved):

tuliomir:
Hello @julia-eileen ,
We have a rule in this repository that requires commits made to it to be signed before merging a PR. Could you please sign your commits so that this branch complies with the rule?
Here are some resources to help with that:
- GitHub on [signing commits](https://docs.github.com/authentication/managing-commit-signature-verification/about-commit-signature-verification)
- A single command for [signing multiple commits](https://superuser.com/a/1123928), via SO

If you want, you can also squash all commits into one, so that the signing is made easier.

julia-eileen:
> Hello @julia-eileen , We have a rule in this repository that requires commits made to it to be signed before merging a PR. Could you please sign your commits so that this branch complies with the rule? Here are some resources to help with that:
> 
>     * GitHub on [signing commits](https://docs.github.com/authentication/managing-commit-signature-verification/about-commit-signature-verification)
> 
>     * A single command for [signing multiple commits](https://superuser.com/a/1123928), via SO
> 
> 
> If you want, you can also squash all commits into one, so that the signing is made easier.

Hi @tuliomir ,

Thank you for pointing that out and for providing the helpful guide! I’ve followed the instructions, signed all of my commits. Please let me know if everything looks good now or if there’s anything else I need to adjust. Thanks again for your support and patience! 🦊

PR #338, opened 2024-10-24
feat: revamp tokens UI

julia-eileen:
### Acceptance Criteria
- This PR includes the revamp of the following screens:
  - See All - Tx & Blocks
  - Tokens List
  - Tokens Balance
- Since many of these screens were originally written using class component definitions, we had to duplicate the class files, adopting the naming convention `NewUi<component_name>.js`, as we couldn't access the feature flag hook inside them. The logic for determining which files to load is handled in the higher-level functional component.
- We made some styling adjustments on the home page to closely match the Figma design.
- The fix for the `li` warning on the home page is still pending. This refactor will be included in the next PR.

### Security Checklist
- [ ] Ensure that no new dependencies were added unless strictly necessary, and no development dependencies were included as production ones, as this could increase the risk of dependency hijacking.

https://github.com/user-attachments/assets/cdf955d6-18cf-4e19-b37a-76d379915939

https://github.com/user-attachments/assets/1915d8f0-4be6-430a-9114-8e4c8766a3f2

https://github.com/user-attachments/assets/3455d6fd-ba41-4ef8-b3eb-ae7cd184c053

https://github.com/user-attachments/assets/fb9998eb-853c-4975-9781-4047a2f34d81

https://github.com/user-attachments/assets/48d9070d-e75f-4c15-bd7b-d778f488293f

https://github.com/user-attachments/assets/33803327-12d7-4f62-974f-a9b8391699fb


src/components/NewUiAddressDetailExplorer.js
```diff
@@ -0,0 +1,644 @@
...
+import { isEqual, find } from 'lodash';
+import metadataApi from '../api/metadataApi';
+import addressApi from '../api/addressApi';
+import txApi from '../api/txApi';
+import ErrorMessageWithIcon from '../components/error/ErrorMessageWithIcon';
+import NewUiAddressHistory from './NewUiAddressHistory';
+import NewUiAddressSummary from './NewUiAddressSummary';
+import Spinner from './Spinner';
+
+class NewUiAddressDetailExplorer extends React.Component {
```
tuliomir, line 22: question: In #317 this component was rewritten using functional components, which would allow using the hooks needed. Please use that component to reuse the complex logic and add only the new interface elements.

src/components/token/NewUiTokenAutoCompleteField.js
```diff
@@ -0,0 +1,245 @@
...
+import { constants as hathorLibConstants } from '@hathor/wallet-lib';
+import { connect } from 'react-redux';
+
+const DEBOUNCE_SEARCH_TIME = 200; // ms
+
+const mapStateToProps = state => {
+  return { serverInfo: state.serverInfo };
+};
+
+class NewUiTokenAutoCompleteField extends React.Component {
```
tuliomir, line 14: question: This duplication of code raises some worries here, because we will soon need to refactor those components to functional. This unification of the logic between files - even though it's supposed to be the same logic - can generate accidental bugs.

Would it be possible for us to find another way of making this distinction between the new UI and the old UI without using hooks? That way, we could have the same `renderUi()` / `renderNewUi()` pattern implemented in those components.

A first possibility that occurs to me is to create a new Redux property storing this new UI information and use it in the legacy class components. That property could be updated by the hook itself, since it's running correctly at the higher level components.

Do you think this is a valid solution, or maybe would you have other ideas on how to implement an alternative?

tuliomir (changes requested):

src/components/NewUiSortableTable.js
```diff
@@ -0,0 +1,158 @@
...
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+import PropTypes from 'prop-types';
+import Spinner from './Spinner';
+import ErrorMessageWithIcon from './error/ErrorMessageWithIcon';
+
+class NewUiSortableTable extends React.Component {
```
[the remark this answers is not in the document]
julia-eileen: We haven’t changed the behavior of these components; instead, we replicated them specifically to adjust styles for the revamp (the screen we’re working on uses SortableTable, which needed a new css class). This approach allows us to keep the old UI until the revamp is complete, as feature flags aren’t accessible within class components. Once the revamp is finished, the plan is to fully replace the old UI with the new version. We're not introducing any new class components—only maintaining existing ones as they are.

Refactoring to functional components may present challenges, as it limits our ability to modify these files and restricts further refactoring of completed work.

src/components/CustomSelect.js
```diff
@@ -0,0 +1,53 @@
...
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React, { useEffect, useRef, useState } from 'react';
+import { ReactComponent as RowDown } from '../assets/images/arrow-down-nav-dropdown.svg';
+
+const CustomSelect = ({ options, onSelect, value }) => {
```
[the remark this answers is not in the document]
julia-eileen: The component has been renamed to HathorSelect, and docstrings have been added to explain the expected structure of the options prop, as well as the purpose of each function and parameter.

Let me know if it's well structured now 🦊.

src/components/NewHathorAlert.js
```diff
@@ -0,0 +1,37 @@
...
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+import { ReactComponent as SuccessIcon } from '../assets/images/success-icon.svg';
+
+class NewHathorAlert extends React.Component {
```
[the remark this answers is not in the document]
julia-eileen: Refactored it to a functional component, please review.

src/components/NewUiAddressDetailExplorer.js
julia-eileen: Removed to use the refactored functional component. Please review.

src/screens/AddressDetail.js
```diff
@@ -27,7 +30,11 @@ const AddressDetail = () => {
   }
 
   if (latestMode) {
-    return <AddressDetailExplorer history={history} match={{ params }} />;
+    return newUiEnabled ? (
+      <NewUiAddressDetailExplorer history={history} match={{ params }} />
+    ) : (
+      <AddressDetailExplorer history={history} match={{ params }} />
```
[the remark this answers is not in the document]
julia-eileen: Removed to use the refactored functional component. Please review.

src/components/tx/Transactions.js
```diff
@@ -278,17 +284,110 @@ function Transactions({ shouldUpdateList, updateData, title }) {
...
+      </div>
+    );
+  };
+
+  const loadNewPagination = () => {
+    if (transactions.length === 0) {
+      return null;
+    }
+
+    return !noPagination ? (
```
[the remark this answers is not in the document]
julia-eileen: Fixed as sugested. Please review.

src/components/token/NewUiTokenAutoCompleteField.js
julia-eileen: I share the concern about potential issues with code duplication, especially considering future refactoring to functional components. Here’s an approach we could take to simplify this while keeping the same `renderUi` and `renderNewUi` pattern:

- We can maintain a single file for the class components.
- The higher-level component, which is functional and has access to the feature flag, could pass down a prop to the class component, indicating whether to render the new or old UI.
- This way, the class component can determine the appropriate UI to render without needing hooks directly, preserving a single source of truth and reducing the risk of accidental bugs due to duplicated logic.

Let me know if this approach seems feasible or if you have any additional thoughts!

tuliomir: That looks like a great solution to me :+1: . We can move on with this approach.

src/components/NewUiSortableTable.js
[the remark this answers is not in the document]
julia-eileen: I've refactored it to meet the requested changes. Please review when you have a chance

src/components/token/NewUiTokenAutoCompleteField.js
julia-eileen: Just a quick note: since some components rely on inheritance, I wasn't able to create new methods specifically for the new UI. Instead, I adjusted the content conditionally based on the feature flag.

src/components/token/TokenBalances.js
```diff
@@ -366,6 +370,51 @@ function TokenBalances({ maintenanceMode }) {
...
+
+  const renderNewUi = () => (
+    <div className="container-title-page">
+      <p className="title-page">Token Balance</p>
+      {renderSearchField()}
+
+      <div className="token-balances-information-wrapper">
+        <h1>
+          {this.state.name === undefined
+            ? 'Hathor - HTR'
```
r4mmer, line 382: `Hathor - HTR` should not be hardcoded.
We have a `state.serverInfo.native_token` which has a name and symbol that should be used here.

src/components/token/TokenRow.js
```diff
@@ -22,7 +26,48 @@ function TokenRow({ token }) {
...
+        {idStart}
+        <div className="ellipsis">
+          <div className="ellipsi"></div>
+          <div className="ellipsi"></div>
+          <div className="ellipsi"></div>
+        </div>
+        {idEnd}
+      </div>
+    );
+  };
```
r4mmer, line 48: Didn't you create a component for this?
Should we use it here?

src/components/tx/TxRow.js
```diff
@@ -57,8 +39,10 @@ const TxRow = ({ tx }) => {
 
   const renderNewUi = () => (
     <tr onClick={_e => handleClickTr(tx.tx_id)}>
-      <td className=" d-lg-table-cell pe-3">{renderIdCell(tx.tx_id)}</td>
-      <td className=" d-lg-table-cell pe-3">{parseTimestamp(tx.timestamp)}</td>
+      <td className=" d-lg-table-cell pe-3">{ellipsis ? renderIdCell(tx.tx_id) : tx.tx_id}</td>
```
r4mmer, line 42: Maybe we should use the EllipsiCell component here as well

src/components/AddressHistory.js
```diff
@@ -160,6 +173,77 @@ class AddressHistory extends SortableTable {
...
+          <td className="pe-3">{hathorLib.transactionUtils.getTxType(tx)}</td>
+          <td className="pe-3">
+            <EllipsiCell id={tx.tx_id} />
+          </td>
+          <td className="pe-3 td-mobile date-cell">
+            {dateFormatter.parseTimestampNewUi(tx.timestamp)}
+          </td>
+          <td className="state td-mobile">{statusElement}</td>
+          <td className="value td-mobile">
+            <span style={{ color: tx.balance < 0 ? '#991300' : '#44A32E' }}>{prettyValue}</span>
```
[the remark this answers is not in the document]
julia-eileen: Fixed as sugested. Please review when you can.

src/components/NewHathorAlert.js
```diff
@@ -0,0 +1,64 @@
...
+
+import React, { useRef, forwardRef, useImperativeHandle } from 'react';
+import { ReactComponent as SuccessIcon } from '../assets/images/success-icon.svg';
+
+/**
+ * NewHathorAlert - A functional alert component with an icon and customizable display duration.
+ * This component can be shown for a specified duration by calling the `show` method from its parent component.
+ *
+ * @param {Object} props - Component properties.
+ * @param {string} props.type - Defines the alert type for styling (e.g., "success", "error").
```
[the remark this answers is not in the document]
julia-eileen: We haven’t defined it yet because, until this PR, we’ve only encountered the success case. Once we come across other scenarios in Figma, we can define them as needed. This way, we align the development flow closely with Figma and avoid overthinking scenarios we haven’t encountered yet. Could it be this way?

src/components/token/TokenBalances.js
julia-eileen: How can I access the `serverInfo` object here? Since we're now in the refactored functional component, we don't have access to the previous state, and none of the defined state variables reference serverInfo.

src/components/token/TokenRow.js
julia-eileen: Indeed. Fixed as sugested.

src/components/token/TokenBalances.js
julia-eileen: Nevermind, it's available through the redux store.

src/components/NewHathorAlert.js
```diff
@@ -0,0 +1,64 @@
...
+ * @param {string} props.text - The message text displayed in the alert.
+ * @param {React.Ref} ref - A reference to call the `show` method from parent components.
+ *
+ * Usage:
+ * ```javascript
+ * const alertRef = useRef(null);
+ * alertRef.current.show(2000); // Show the alert for 2 seconds
```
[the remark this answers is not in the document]
julia-eileen: fixed as sugested.

src/components/NewHathorAlert.js
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * NewHathorAlert - A functional alert component with an icon and customizable display duration.
+ * This component can be shown for a specified duration by calling the `show` method from its parent component.
+ *
+ * @param {Object} props - Component properties.
+ * @param {string} props.type - Defines the alert type for styling (e.g., "success", "error").
+ * @param {string} props.text - The message text displayed in the alert.
+ * @param {React.Ref} ref - A reference to call the `show` method from parent components.
+ *
+ * Usage:
```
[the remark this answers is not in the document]
julia-eileen: fixed as sugested.

src/components/token/TokenBalances.js
julia-eileen: Fixed as sugested. Please review when possible.

src/components/tx/TxRow.js
julia-eileen: Agreed. Fixed as sugested.

src/components/token/Tokens.js
tuliomir, line 1: chore(non-blocking): When navigating to `/tokens`, I receive the following message on console:
```
Warning: Failed prop type: The prop `title` is marked as required in `Tokens`, but its value is `undefined`.
```

This is, however, outside of the stope for this PR. I just want to register it here for future reference.

tuliomir (approved):

src/components/token/TokenBalances.js
```diff
@@ -10,17 +10,21 @@ import PropTypes from 'prop-types';
 import { get, last, find, isEmpty } from 'lodash';
 import { useHistory } from 'react-router-dom';
 import { numberUtils, constants as hathorLibConstants } from '@hathor/wallet-lib';
+import { useNewUiEnabled } from '../../hooks';
 import TokenBalancesTable from './TokenBalancesTable';
 import tokensApi from '../../api/tokensApi';
 import PaginationURL from '../../utils/pagination';
 import ErrorMessageWithIcon from '../error/ErrorMessageWithIcon';
 import TokenAutoCompleteField from './TokenAutoCompleteField';
+import { useSelector } from 'react-redux';
```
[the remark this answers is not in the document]
julia-eileen: Fixed as sugested, thanks.

src/components/token/TokenBalances.js
```diff
@@ -366,6 +379,51 @@ function TokenBalances({ maintenanceMode }) {
...
+  const renderNewUi = () => (
+    <div className="container-title-page">
+      <p className="title-page">Token Balance</p>
+      {renderSearchField()}
+
+      <div className="token-balances-information-wrapper">
+        <h1>
+          {tokenName === undefined
+            ? `${serverInfo.native_token.name} (${serverInfo.native_token.symbol})`
+            : `${tokenName} (${tokenSymbol})`}
```
r4mmer, line 392: I think this changed from `name - symbol` to `name (symbol)`.
If this was intended you can disregard this comment.

r4mmer (approved):

src/components/token/TokenBalances.js
julia-eileen: Yes, it was intentional to align with the Figma prototype design.

PR #342, opened 2024-11-13
refactor: centralize version number on package

tuliomir:
### Acceptance Criteria
- All code paths that require the version number should fetch if from the `package.json` directly
- The `check_version` script should validate both application version numbers from the lockfile

### Security Checklist

### DIFF
--- a/scripts/check_version
+++ b/scripts/check_version
@@ -1,16 +1,26 @@
 #!/bin/bash
 
-SRC_VERSION=`grep "const VERSION " ./src/constants.js | cut -d"'" -f2`
+set -e  # Exit on any command failure.
+set -u  # Exit on unset variables.
+
 PACKAGE_VERSION=`grep '"version":' ./package.json | cut -d '"' -f4`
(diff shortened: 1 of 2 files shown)


luislhl (approved):

luislhl (approved):

scripts/check_version
```diff
@@ -1,17 +1,27 @@
 #!/bin/bash
 
-SRC_VERSION=`grep "const VERSION " ./src/constants.js | cut -d"'" -f2`
+set -e  # Exit on any command failure.
+set -u  # Exit on unset variables.
+
 PACKAGE_VERSION=`grep '"version":' ./package.json | cut -d '"' -f4`
```
r4mmer, line 6: Why not also use the node command here?

r4mmer (approved):

scripts/check_version
tuliomir: We can transform this into a common action later, centralizing the code for the version check for all our NodeJs applications.

PR #352, opened 2024-12-17
fix: Error handling on nav search for new UI

tuliomir:
When the user tries to use the search box with the new UI and inputs invalid data ( ex.: an address from another network ), the app throws a javascript error that the `ref` does not exist.

This was solved by providing an error handler for the Navigation element, with the same fixed positioning as the current UI. Screenshots below for reference.

### Acceptance Criteria
- The correct error message is shown when the search box has invalid content

> [!WARNING]
> The PR indicates there are hundreds of changed lines. This happened because I needed to add one indentation to the Navigation file. To avoid confusion on the review proccess, I recommend activating the **Hide Whitespaces** feature on the files tab.
> That way only the actual changes are displayed, and not the hundreds of indentation changes.

#### New UI
New UI

#### Current UI
Current UI

### Security Checklist

### DIFF
--- a/src/components/Navigation.js
+++ b/src/components/Navigation.js
@@ -29,6 +29,7 @@ import {
   REACT_APP_NETWORK,
 } from '../constants';
 import { toggleTheme } from '../actions';
+import NewHathorAlert from './NewHathorAlert';
 
 function Navigation() {
   const history = useHistory();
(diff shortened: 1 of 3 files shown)


pedroferreira1 (approved):

src/components/Navigation.js
```diff
@@ -144,132 +198,89 @@ function Navigation() {
...
-        <Sidebar close={() => setShowSidebar(false)} open={showSidebar} />
-      </nav>
+          </div>
+          <Sidebar close={() => setShowSidebar(false)} open={showSidebar} />
+        </nav>
+        <NewHathorAlert
+          ref={alertErrorRef}
+          text="Invalid hash format or address"
+          type="error"
+          fixedPosition
```
r4mmer, line 281: I think this is `undefined` if its not being initialized

r4mmer (approved):

PR #378, opened 2025-01-16
fix: Alert unmount and overlay

tuliomir:
During some tests where many components were being clicked in short sequence, the `NewHathorAlert` component could break by being unable to manipulate the dom correctly: the expected element would already be unmounted when the timer finished.

Also, the alert container for fixed positioning, used in the Navigation search, was overlaying the screen elements, preventing clicks on the application elements.

### Acceptance Criteria
- The new alert component should not break under race conditions
- The new alert component should not prevent click in any screen elements

### Security Checklist

### DIFF
--- a/src/components/NewHathorAlert.js
+++ b/src/components/NewHathorAlert.js
@@ -5,7 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import React, { useRef, forwardRef, useImperativeHandle, useEffect } from 'react';
+import React, { useRef, forwardRef, useImperativeHandle } from 'react';
 import { ReactComponent as SuccessIcon } from '../assets/images/success-icon.svg';
 
 /**
@@ -15,7 +15,7 @@ import { ReactComponent as SuccessIcon } from '../assets/images/success-icon.svg
  * @param {Object} props - Component properties.
  * @param {string} props.type - Defines the alert type for styling (e.g., "success", "error").
  * @param {string} props.text - The message text displayed in the alert.
- * @param {boolean} [props.fixedPosition=false] - If true, the alert will be fixed at the bottom right of the screen
+ * @param {boolean} [props.fixedPosition=false] - If true, the alert will be fixed at the bottom center of the screen
  * @param {React.Ref} ref - A reference to call the `show` method from parent components.
  *
  * @example:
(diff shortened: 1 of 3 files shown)


src/newUi.scss
```diff
@@ -975,12 +979,6 @@ nav {
   justify-content: space-between;
   z-index: 9999;
   margin: 0;
-
-  &.fixed-position {
-    position: fixed !important;
-    right: 15%;
-    bottom: 50%;
-  }
```
tuliomir, line 983: Removing CSS leftover from #352, obsolete since #377

src/components/NewHathorAlert.js
```diff
@@ -34,41 +35,80 @@ const NewHathorAlert = forwardRef(({ type, text, showAlert, fixedPosition }, ref
...
+   * Renders the alert component
+   * @returns {Element}
+   */
+  function renderAlertDiv() {
+    return (
+      <div
+        ref={alertDiv}
+        className={`new-hathor-alert alert alert-${type} alert-dismissible fade`}
+        role="alert"
+        style={{ display: 'flex', flexDirection: 'row' }}
```
tuliomir, line 93: I tried to avoid this `style` from the previous implementation, but the fade animations just wouldn't work without it. Decided to leave this cleanup to a future refactor.

andreabadesso (approved):

src/components/NewHathorAlert.js
```diff
@@ -34,41 +35,80 @@ const NewHathorAlert = forwardRef(({ type, text, showAlert, fixedPosition }, ref
...
+
       alertDiv.current.classList.add('show');
-    } else {
-      alertDiv.current.classList.remove('show');
+      setTimeout(() => {
+        // By the time the timeout is called, the component may have been unmounted
+        if (alertDiv?.current) {
+          alertDiv.current.classList.remove('show');
+        }
+      }, duration);
```
pedroferreira1, line 75: You don't have this `duration` parameter here, if I'm not wrong. The diffs are a bit confusing but it seems like you should be receiving it as a parameter in the `showAlertComponent` also.

pedroferreira1 (approved):

src/components/NewHathorAlert.js
tuliomir: This function is actually declare inside the `show()` function, so it shares the `duration` parameter.
The diff visualizarion is confusing, since some of the methods were obsolete and were removed, but you can see it more clearly on the [file itself](https://github.com/HathorNetwork/hathor-explorer/blob/bc594aee5485f6e9c262e63b3c403edee1ec981e/src/components/NewHathorAlert.js#L63-L75).

PR #385, opened 2025-02-17
feat: add blueprint list

pedroferreira1:
### Acceptance Criteria
- Add new Navigation for nano
- Add blueprint list with two tabs (on chain and built in)
- Refactor EllipsisCell component to receive the elements count as props because now we have different counts depending on the component
- Fix missing arrow in sorting header for Tokens table

https://github.com/user-attachments/assets/482e3010-a03c-4050-93ac-e655f9603c94

https://github.com/user-attachments/assets/1fab628f-7f16-4d43-a784-44f58079c499

### Security Checklist

### DIFF
--- a/src/components/SortableTable.js
+++ b/src/components/SortableTable.js
@@ -21,7 +21,7 @@ class SortableTable extends React.Component {
 
   renderTable(content) {
     return this.props.newUiEnabled ? (
-      <table className="table-stylized table-tokens" id="">
+      <table className={`table-stylized ${this.props.tableClass ? this.props.tableClass : 'table-tokens'}`} id="">
         {content}
       </table>
     ) : (
@@ -150,7 +150,7 @@ class SortableTable extends React.Component {
                   if (this.props.calculatingPage) {
                     return 'page-item disable-button';
                   }
-                  return 'page-item active';
+                  return 'page-item';
                 })()}
               >
                 <button onClick={e => this.props.onNextPageClicked(e)} className="page-link">
@@ -196,7 +196,7 @@ class SortableTable extends React.Component {
  * order: If sorted field must be ordered asc or desc
  * tableHeaderClicked: This indicates that user wants data to be sorted by a determined field
  * calculatingPage: Indicates if next page is being retrieved from explorer-service
- * tableClasses: Extra classes to add to the table element
+ * tableClass: CSS class to add to the table element. The default value is 'table-tokens'.
  */
 SortableTable.propTypes = {
   data: PropTypes.array.isRequired,
@@ -209,6 +209,7 @@ SortableTable.propTypes = {
   tableHeaderClicked: PropTypes.func,
   sortBy: PropTypes.string,
   order: PropTypes.string,
+  tableClass: PropTypes.string,
 };
 
 export default SortableTable;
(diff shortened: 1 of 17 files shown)


src/components/SortableTable.js
```diff
@@ -150,7 +155,7 @@ class SortableTable extends React.Component {
                   if (this.props.calculatingPage) {
                     return 'page-item disable-button';
                   }
-                  return 'page-item active';
+                  return 'page-item';
```
pedroferreira1, line 158: Ana decided to remove the active style from table pagination buttons, so both pagination buttons have the same style now.

tuliomir (approved):

tuliomir (approved):

src/components/nano/OnChainBlueprintsTable.js
```diff
@@ -0,0 +1,76 @@
+import React from 'react';
+import SortableTable from '../SortableTable';
+import EllipsiCell from '../EllipsiCell';
+import dateFormatter from '../../utils/date';
+
+// XXX We should use function component with SortableTable as a component
+// but renderTableHead and renderTableBody are implemented and not
+// expected as a props, so it demands a bigger refactor
+class OnChainBlueprintsTable extends SortableTable {
```
r4mmer, line 9: suggestion(non-blocking): Maybe it shouldn't extend SortableTable but use it

src/components/SortableTable.js
```diff
@@ -21,7 +21,12 @@ class SortableTable extends React.Component {
 
   renderTable(content) {
     return this.props.newUiEnabled ? (
-      <table className="table-stylized table-tokens" id="">
+      <table
+        className={`table-stylized ${
+          this.props.tableClass ? this.props.tableClass : 'table-tokens'
+        }`}
+        id=""
```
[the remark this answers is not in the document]
pedroferreira1: I don't think so, removed it https://github.com/HathorNetwork/hathor-explorer/pull/385/commits/385ac89ce8625b25e839d3403f934513eaf18f88

src/components/nano/OnChainBlueprintsTable.js
pedroferreira1: We need to extend it because we need to override the `renderTableBody` and `renderTableHead` methods. They should have been designed as props, so we could just use the component with the props defined but it wasn't.

That's the comment in the file, we should refactor it to receive these methods as props and change all the inheritance to use it.

src/utils/pagination.js
```diff
@@ -42,18 +42,46 @@ class PaginationURL {
...
+  setURLParameters(data, paramsToDelete = []) {
     const url = new URL(window.location.href);
     for (const param in data) {
       if (param in this.parameters) {
         url.searchParams.set(param, data[param]);
       }
     }
+
+    for (const param of paramsToDelete) {
+      if (!url.searchParams.get(param)) {
```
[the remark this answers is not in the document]
pedroferreira1: It makes sense and the method `has` exists. Changed it https://github.com/HathorNetwork/hathor-explorer/pull/385/commits/548f41d3c3f6e1cd113c7384d53a74052a8e51a8

PR #388, opened 2025-03-03
chore: improve inputs and outputs lists

obiyankenobi:
### Acceptance Criteria
- add new design for inputs and outputs list;
- make whole title line of dropdown element clickable to improve usability;
- compact hash format (like in the initial page) changed to font-size 16px;
- hash item list (parents & children) font-size to 14px;

### Security Checklist

### Screenshots
image


src/components/tx/TxData.js
```diff
@@ -367,7 +362,8 @@ class TxData extends React.Component {
     const renderOutputLink = idx => {
       if (idx in this.props.spentOutputs) {
         return (
-          <span>
+          // TODO for some reason, bold does not work with SF Pro
```
obiyankenobi, line 365: @pedroferreira1 @tuliomir any idea why? I tried all options (bold, strong, 700, etc). I believe all text we have as bold on the website (that's currently working) uses Mona Sans font, not SF Pro.

src/newUi.scss
```diff
@@ -1128,17 +1128,19 @@ nav {
...
 }
 
+.fs-14 {
+  font-size: 14px !important;
+}
+
 .table-details td {
-  color: var(--purple-text);
-  padding: 20px 10px;
+  padding: 20px 0px;
```
obiyankenobi, line 1136: Removed the padding from the lists. It's not according to Ana's design, but I personally think it looks better.

Original
image


New
image

src/newUi.scss
```diff
@@ -2343,7 +2351,7 @@ nav {
   flex-direction: row;
   gap: 2px;
   align-items: center;
-  font-size: 14px;
+  font-size: 16px;
```
obiyankenobi, line 2354: The original design has both this and the timestamp column as 14px. However, we currently use 14px for this and 16px for the timestamp. Since both can use 16px, it's better for usability.

src/components/tx/TxData.js
```diff
@@ -509,14 +505,19 @@ class TxData extends React.Component {
       ));
     };
 
-    const renderNewUiDivList = hashes => {
+    const renderTxListWithSpacer = hashes => {
+      const obj = hashes.map( h => <Link className="fs-14" to={`/transaction/${h}`}>{h}</Link> );
```
obiyankenobi, line 509: Increased font size to 14px to improve usability. It was too small before.

Original
image


New
image

src/components/DropDetails.js
```diff
@@ -29,11 +29,10 @@ export const DropDetails = ({ title, startOpen, onToggle, children }) => {
 
   return (
     <div className={`container-drop-div ${open ? 'container-drop-div-open' : ''}`}>
-      <div className="container-drop-header">
+      <div className="container-drop-header" onClick={click}>
```
tuliomir, line 32: note(non-blocking): This behavior was supposed to be fixed in #378, but an issue with overlays postponed it.

I'm sharing this here because I believe there are still some places where this kind of overlay issue is still happening, and in these cases the `container-drop-header` click event will seem not to be working, breaking the manual QA.

I couldn't find precisely _where_ to open an objective issue about it, but will do it soon.

src/components/tx/TxData.js
tuliomir: Based on our recent conversation, it seems the bold font files were available on the developer machines that implemented these changes ( and also on mine, as I couldn't reproduce the error ), but were not available in the repository for all users.

Adding those to the repository will probably solve this issue.

obiyankenobi: Fixed by adding the font file (`src/assets/fonts/SFPRODISPLAYBOLD.OTF`) and definition on scss.

src/components/tx/TxData.js
```diff
@@ -820,25 +819,22 @@ class TxData extends React.Component {
         }
         return <Link to={`/token_detail/${token.uid}`}>{token.uid}</Link>;
       };
-      const tokens = this.state.tokens.map(token => {
-        return (
-          <div key={token.uid}>
+      const obj = this.state.tokens.map(token => (
+        // TODO I don't think we have a TokenMarker here on Figma. Remove?
```
obiyankenobi, line 823: @tuliomir is that the case? Should we remove TokenMarkers?

src/components/tx/TxData.js
```diff
@@ -820,26 +820,21 @@ class TxData extends React.Component {
...
             <span>
-              {token.name} <strong>({token.symbol})</strong> | {renderTokenUID(token)}
+              {token.name} ({token.symbol})
             </span>
           </div>
-        );
-      });
-      return this.props.newUiEnabled ? (
-        <DropDetails title="Tokens">{tokens}</DropDetails>
-      ) : (
```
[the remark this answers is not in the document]
obiyankenobi: Correct. I talked to @pedroferreira1 and he said we should already disable the old UI. All screens being added are not taking the old design into account anymore.

src/components/DropDetails.js
obiyankenobi: I don't understand. This makes the whole dropdown header row clickable. Is it supposed to be fixed in #378?

tuliomir: We currently have some "ghost" divs with a high z-index that block the click behavior of some elements on the screen.

This is the issue that was partially solved in 378. However, every now and then I try to click some element on screen and notice it's not working. Just a small scroll up or down is enough to avoid this problem.

During an eventual QA, this issue might be mistaken for a bug on your code on the click behavior of the header row, and that's why I brought it here: to clarify the actual source of the problem.

PR #386, opened 2025-02-24
feat: Upgrades `react-scripts` to v5

tuliomir:
### Motivation
This repository was still using React Scripts v3 and the upgrade to v4 was facing problems complicated enough that it was not being worth the effort to add this intermediary step. So, a decision was made to jump directly to v5.

This decision considered the tradeoffs discussed in https://github.com/HathorNetwork/hathor-wallet/issues/441 and took advantage of know-how acquired in https://github.com/HathorNetwork/hathor-wallet/pull/712 to greatly decrese the effort needed for this upgrade.

### Acceptance Criteria
- Upgrades `react-scripts` from v3 to v5

### Security Checklist
- [X] Make sure you do not include new dependencies in the project unless strictly necessary and do not include dev-dependencies as production ones. More dependencies increase the possibility of one of them being hijacked and affecting us.

PR #404, opened 2025-04-24
fix: `cjs` file handling for axios

tuliomir:
### Summary
We were having errors when trying to load the "Transaction Details" screen for a nano contract tx. This was happening because the module for Axios was not being found by the browser, and instead if was being fed only a string containing the module path.

A fix had to be implemented in the `config-override.js` file to instruct WebPack v5 on where to find the module, which is in a file format we did not configure previously on the CRA upgrade ( #386 ).

### Acceptance Criteria
- The "Nano Contract Overview" section of a nc transaction should be exhibited on the "Transaction Details" screen

### Security Checklist

### DIFF
--- a/config-overrides.js
+++ b/config-overrides.js
@@ -27,12 +27,18 @@ module.exports = function override(config) {
   };
 
   // Relaxing js/mjs extension resolve
-  config.module.rules.push({
-    test: /\.m?js/, // Apply this rule to .js and .mjs files
-    resolve: {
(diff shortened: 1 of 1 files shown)


pedroferreira1 (approved):

luislhl (approved):
Is there a way to write a test that would have caught this problem?

If so, maybe we could write one in a next PR.

tuliomir:
For now, we only run the manual tests after starting the release process in the `release-candidate` and `release` branches. But I agree this would be an important automated test to have on the CI.

PR #446, opened 2025-10-27
feat: decode nc-fail voided hex to show in the tx data component

pedroferreira1:
### Acceptance Criteria
- Decode nc-fail to improve UX for user

After

image

Before

image

### Security Checklist

### DIFF
--- a/src/components/tx/TxData.js
+++ b/src/components/tx/TxData.js
@@ -18,6 +18,7 @@ import TokenMarkers from '../token/TokenMarkers';
 import TxAlerts from './TxAlerts';
 import TxMarkers from './TxMarkers';
 import dateFormatter from '../../utils/date';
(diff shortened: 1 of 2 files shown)


src/components/tx/TxData.js
```diff
@@ -466,22 +467,24 @@ class TxData extends React.Component {
       if (hashes.length === 0) {
         return null;
       }
-      if (hashes.length === 1) {
-        const h = hashes[0];
+      const ncFailedExecution = 'Nano contract failed execution';
+      const renderSingleElement = h => {
+        if (h === NANO_CONTRACT_EXECUTION_FAIL) {
+          return <span>{ncFailedExecution}</span>;
```
tuliomir: suggestion: You could inline `ncFailedExecution` here, as it's the only place it's used..

PR #519, opened 2026-06-17
test(e2e): add Playwright navigation-smoke suite against mainnet

raul-oliveira:
### Description

Adds a Playwright E2E suite for the explorer (there were none) plus a first navigation-smoke journey. Modeled on the `hathor-rpc-lib-e2e` architecture, minus the wallet/MetaMask/Snap layers (the explorer is read-only). Playwright boots the explorer's own CRA dev server locally and runs it against the real public **mainnet** backends with real Unleash; the read-only, stateless app runs fully parallel.

### Acceptance Criteria

- E2E harness in place: Playwright config (mainnet env override), scoped `tests/e2e/tsconfig.json`, conventions doc (`tests/e2e/e2e.md`), immutable anchors, page objects and a page-scoped fixture.
- `e2e` / `e2e:headed` / `e2e:ui` npm scripts added; `playwright-report/`, `test-results/`, `reports/screenshots/` gitignored.
- Navigation smoke covers all 11 main routes (`/`, `/transactions`, `/blocks`, `/tokens`, `/token_balances`, `/statistics`, `/dag`, `/features`, `/network`, `/decode-tx`, `/push-tx`): each renders its identifying element with no global error state, capturing a per-route evidence screenshot.
- Asserts the mainnet env override took effect, top-nav click navigation and footer presence.
- Suite is green: 14/14 passing against real mainnet.
- The test browser launches with `--disable-web-security` (the real backends only allow the production origin); scoped to the test browser, nothing in the app changes. Cleaner alternatives are recorded in the design doc.

### Security Checklist
- [x] Make sure you do not include new dependencies in the project unless strictly necessary and do not include dev-dependencies as production ones. The only new dependencies are test-only devDependencies (`@playwright/test`, `typescript`, `@types/node`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/superpowers/plans/2026-06-16-explorer-e2e-navigation-smoke.md
+++ b/docs/superpowers/plans/2026-06-16-explorer-e2e-navigation-smoke.md
@@ -1,0 +1,801 @@
+# Explorer E2E — Infrastructure & Navigation Smoke Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+> **⚠️ GIT POLICY (user override):** This user FORBIDS the agent from running `git add`,
+> `git commit`, or `git stash`. Every "Commit" step below shows the command for the **user to
+> run manually**. At each commit checkpoint, STOP and ask the user to run it — never run it
+> yourself.
+
+**Goal:** Stand up a Playwright E2E suite for hathor-explorer (config, fixture, page objects,
+conventions doc, anchors) and ship a navigation-smoke journey that asserts every main route
+renders its identifying content with no global error, against real public mainnet backends.
+
+**Architecture:** Playwright boots the explorer's own CRA dev server locally (`webServer`),
+pointed at real public **mainnet** backends via `webServer.env` (OS env wins over the
+committed `.env.local`). The app is read-only and stateless → `fullyParallel: true`. Tests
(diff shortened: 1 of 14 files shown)


docs/superpowers/plans/2026-06-16-explorer-e2e-navigation-smoke.md
```diff
@@ -0,0 +1,801 @@
+# Explorer E2E — Infrastructure & Navigation Smoke Implementation Plan
```
tuliomir, line 1: thought(non-blocking): Reconsider adding the plan and spec into the public repo.

My impression is that they add no more value to the future of the application than what a PR and squash commit already do on the permanent git log history.

tests/e2e/tsconfig.json
```diff
@@ -0,0 +1,15 @@
+{
+  "compilerOptions": {
+    "target": "ES2021",
```
tuliomir, line 3: question(non-blocking): Does this target allow `BigInt`s? They were added only on NodeJS 22, so maybe this can cause some maintenance problems.

We could wait until we find and issue to fix this, though.

package.json
```diff
@@ -63,6 +66,8 @@
   ],
   "devDependencies": {
     "@babel/eslint-parser": "7.26.8",
+    "@playwright/test": "^1.50.1",
+    "@types/node": "^22.13.1",
```
tuliomir, line 70: issue(blocking): Only use pinned versions of dependencies, following the repo pattern

tuliomir (changes requested):

